Issue #42, opened 2020-11-22
Versioning, history, find commits from resource


joepio:
One of the advantages of using and storing Commits, is that we have a full history of every single resource. However, we currently have no way of browsing this information.

## Wishes

- I want to click on a link on a resource page to **see all versions**.
- I want to see a **log of changes** of some resource. This might be different from the versions.
- I want to **see a previous version** of some resource, and I want to **share its URL**.
- I want to **restore a previous version** of some resource, yet keep the previous changes.

## Considerations

- For any list of things (commits for some resource, we should use the Collections model.

### Versions endpoint

`example.com/versions?subject=someresource` lists all the version for the resource included in the Subject query parameter. It returns a collection. [Perhaps first define the Endpoints feature?](https://github.com/ontola/atomic-data/issues/15)

1. User requests a list of versions for some resource
1. A collections is constructed, containing all (or some subset of) the existing versions, one for each Commit. Each version points to a URL of a resource that needs to be constructed (e.g. `example.com/versions?subject=someresource&version=signatureHashBlablabla`)
1. user requests a specific version
1. That version is constructed by the server, using the commits

### Introduce sequential TPF queries

Knowing which commits are linked to a resource, can be done using two TPF queries:

1. Get me all the items where `is-a = commit`
1. AND `subject = mySubject`

Some time ago, the great Ruben Verborgh told me that you can actually convert (all?) SPARQL queries into a set of sequential TPF queries. That's pretty cool. How should we create an interface to facilitate this? It probably requires some form of nesting queries, and some form of combined queries.

But... this kind of discussion quickly turns into a difficult discussion of which methods need to be included in the newly designed query language, and will kind of lead to some almost-turing-complete language, but not fully. Therefore, we might want to try a plugin-like approach, where the query logic resides in code.

### Plugin (endpoint + extend resources)

Atomic Plugins are currently just a bunch of thoughts, but perhaps `Versions` is a logical first plugin. Plugins should be apps that can be added to an Atomic Server at runtime. They might be run using a WASM runtime. In any case, they need to add functionality to an Atomic Server.

Ideally, the Versioning plugin will have:

- An endpoint (see above) for constructing a version of a resource, and listing all versions / commits of a resource 
- Maybe add some links to resources, so you can find the versions of a resource from the resource page.
- A way to restore previous versions (by producing a Commit that reverts some changes) 

## Constructing a version

A Version is a representation of some Resource at some point in time. Constructing one can (theoretically be done in several ways:

[...]

PR #88, opened 2021-01-28
Versioning / history #42

joepio:
#42


joepio:
Still lacks tests and a server implementation

Issue #14, opened 2020-09-17
TPF indexing in Db


joepio:
Although TPF queries are implemented, they are very slow and won't scale - a single TPF query iterates over all individual atoms in the store. To solve this, we need some type of index. Since we're using Sled, a key-value store, we can't use some SQL index, we need to build it ourselves.

One solution is to create two new Sled tree (a new k-v store). In the first one (for searching by Value) every k represents an Atomic Value, and v a vector of all subjects. In the second one for Properties, k = property, v = subject.

However, a very common TPF query will be like this: `* isA SomeClass`. If we only do above indexes, this will still be a costly query, because we'll still iterate over many resources - pretty much _all_ resources will have the `isA` property.

We could improve performance if we'd also store the Property in the `v` fields mentioned above, instead of only storing the subjects.  To prevent unnecessary data duplication / minimize storage impact, it might make sense to not store entire atoms, but to leave out the thing that's already known (the thing in the key).

A TPF query such as `* isA SomeClass` would probably start with using the ValueIndex, which return all SubjectProperty combinations. Then, the implementation will iterate over all SubjectProperties, filtering by property, returning all subjects.

I think Atomic Collections will rely on this query quite a bit: make a list of all Persons (or some class), sorted by some thing. This will do such a TPF query using the indexes, than returns all subjects.

Another possible optimization strategy is caching Collections (which internally use TPF queries). We could rebuild (or invalidate) them on Commits.

joepio:
Resources have an explicit parent. Getting the parent is a cheap operation, but generating the children is expensive, as it currently requires a TPF query that iterates over all items. This specific operation should be very quick, but I'd like to make the optimizations generalizable.

One way to optimize this TPF query, is to keep track of incoming links. We could create a new inverted tree for this: `ValueIndex<valueURL, MentionedIn<propery, Vec<subject>>>`. In human language: you can get a `MentionedIn` for any `value`, which contains a list of properties. For any given property, you can find the subjects that have used that property to refer to the valueURL. 

So when we have this `ValueIndex`, we need to add stuff into it. When do we build the index? I _think_ we should update it whenever we add or remove a resource. We iterate over all the PropVals, check if they have an AtomicURL or ResourceArray, and add the items to the index. When a resource is deleted, it needs to be deleted from the index, too.

joepio:
[...]

PR #153, opened 2021-07-22
#14 value index

joepio:
#14 value index working


joepio:
Before I can merge this, I should have some system that prevents re-creating the cache on every boot.

PR #214, opened 2021-11-19
Autocomplete cli clap

joepio:
I tried to get clap's autocomplete feature to work. Had to shuffle up some files, create a `build.rs` file, but seems to work now.  However, I still have to find out how to publish the generated autocomplete files.


joepio:
Closing this obsolete draft during the requested PR cleanup. It targets master and its old CLI/server restructuring now produces extensive conflicts with develop. Autocomplete remains a useful feature, but should be implemented as a focused change against the current CLI, targeting develop, rather than reviving this historical branch. The branch is retained for reference.

Issue #182, opened 2021-10-03
Check rights of Invite creator


joepio:
We recently introduced Invites #134 

However, the current implementation makes it possible for users to grant themselves rights to edit things that aren't theirs. To prevent this, we need to perform these checks:

- [x] Check the rights of the Inviter on creating the invite (commit)
- [x] Check the rights of the Inviter on redirecting the Invitee
- [ ] #233
- [x] Set expiration date for invites

I think we need to check this _twice_, because after an invite has been created, the rights of the creator could have changed.

PR #234, opened 2021-12-11
#182 invite checks

joepio:
#182


joepio:
Tests, fmt, check, and clippy passing locally.  E2E also.

Issue #114, opened 2021-03-17
Collection caching


joepio:
Collections are currently dynamic resources, which means that they are fully calculated when a user sends a request. That works fine, but it comes at a performance cost, since the DB must be queried. 

How to cache this? How does this interact with the `get_extended_resource` function? How to invalidate the cache? Let's discuss some considerations.

Collections can be sorted and filtered by adding query params. These of course change the dynamic properties such as  `members` and `total_count`. These should be cached seperately.

Since all changes _should be done_ using Commits, we can perform cache invalidations while handling Commits. How does the Commit handler know which resources should be invalidated? For example, let's say I remove the `firstName` property with the `john` value from some `person` Resource. The person first appeared in the collections of people named `john`, but this collection should now be invalidated.

### Invalidate when any attribute of a resource chages

When a Collection iterates over its members, it adds the subject of the collection (including query params) to a K/V `incomingLinks` store where each K is a subject, and each V stands for an array of subjects that link to it. When a commit is applied to resource X, it takes subject X and opens the `incomingLinks` instance of that X. It then proceeds to invalidate all the V items. This will invalidate many collections that could very well result in _exactly the same members_, when re-run. 

### Use TPF index / cache

#14 

If we build an index for all values, most of the expensive part is solve. It just leaves sorting - which still is expensive.

joepio:
Now that a Value Index #14 has been implemented, this issue has become a lot less relevant. Only when collections are _very_ large, it might make sense to cache things like filtering and sorting.

joepio:
I'm noticing that on my laptop, some collections (such as commits) are becoming a bit slower (200 ms for 1400 entries). Makes sense, as all items are iterated and sorted _on the fly_.

That should happen at index time, not at runtime.

joepio:
I've learned a bit more about how to optimise using `sled`, so let's talk about a different implementation idea:

- Add a new `sled::Tree` called `collections_cache` (a KV store) that, for each time a collection is queried, stores an item.
- The `K` is the collection URL (including query params!) + the Value of the value being sorted. This means we always sort by value, and have no indexing cost for sorting. Perhaps we should limit this value to a relatively small amount of characters. If a Collection is not sorted, it is sorted by subject. In that case, use the `subject` as the sortable value.
- The `V` should be a `Vec<String>` of all the subjects in the collections that share this sorted value.
[...]

PR #285, opened 2022-01-15
#114 WIP collection cache

joepio:
TODO:

- [x] Build the cache
- [x] Get rid of `.unwrap()` in `update_member` closures
- [x] Load from cache when possible
- [x] Update CollectionWatcher dynamically

PR Checklist:

- [x] Link to related issue #114
- [x] Add changelog entry linking to issue
- [x] Add tests
- [x] Add tests for authorization
- [x] Add tests for changing data, updating index (e.g. remove stuff)
- [x] More tests?


joepio:
So the thing I'm currently struggling with, is performing the cached query. Currently, `Collection` does not depend on`Db`, which means that I can't access `Db` specific methods from Collection. So should I add collection building methods to `Storelike`?

joepio:
Having some trouble with conditionally reversing the `sled` iterator, depending on the `sort_desc` setting of the user.

```rust
    let iter = if q.sort_desc {
        store.members_index.range(start_key..end_key)
    } else {
        store.members_index.range(start_key..end_key).rev()
    };
```

Fails, because 

```
`if` and `else` have incompatible types
expected type `sled::Iter`
 found struct `std::iter::Rev<sled::Iter>`
```

So I read about using `Box` to prevent this, but that also doesn't work:

```rust
    let iter = if q.sort_desc {
        Box::new(store.members_index.range(start_key..end_key))
    } else {
        Box::new(store.members_index.range(start_key..end_key).rev())
    };
```

Fails:

```
`if` and `else` have incompatible types
expected type `std::boxed::Box<sled::Iter>`
 found struct `std::boxed::Box<std::iter::Rev<sled::Iter>>`rustcE0308
```

joepio:
The changing data test is failing, because the cache is not properly invalidated.

We have this logic, that checks an update atom according to query filters that are being watched:

```
            let should_update = match (&q_filter.property, &q_filter.value) {
                (Some(prop), Some(val)) => prop == &atom.property && val == &atom.value,
                (Some(prop), None) => prop == &atom.property,
                (None, Some(val)) => val == &atom.value,
                // We should not create indexes for Collections that iterate over _all_ resources.
                _ => false,
            };
```

But... That goes wrong when the _sorted_ value differs from the _filtered_ value. So let's say we have a `QueryFilter` that filters by `is_a` `human`, and sorts by `firstname`. Let's say we update the firstname. In this case we want the old Atom to be removed from the index.  So in this case the `should_update` function is called for the new firstname atom. This atom does not contain the `is_a` property or the `human` value. We want to update the existing indexed value, but we can't, because we can't perform the `is_a` `human` check. We need _all_ the properties of the Resource, or we need a pre-confirmed check which collections the resource matches.

We should first check if the `sort_property` matches, then we should check if the resource has the query, I think. This check should be memoized, probably. 

Maybe the problem is more fundamental. The `should_update` function cannot decide if a value needs to be removed, knowing only the Atom and the FilterQuery. 

## Flip it

Say we want to check if an atom will match things. We could _first_ check if the _resource_ matches the query filter. If that is the case, we can continue, and perform the existing checks.

joepio:
So the previous cache invalidation issue has now been solved by passing the _old resource_ to the cache invalidation logic. It is needed to match the `QueryFilter`s in `query_index`.  

Things are working fine if a commit is only applied _once_, but for the `/commit` endpoint, I'm actually applying a Commit _twice_.

joepio:
I think atoms for the `QueryIndex` and the `ReferenceIndex` might need to be a bit different.
The difficult part is `ResourceArrays`.

Let's consider I want add a Paragraph to a Document.
This means I need to add one Atom to the Index.

## ReferenceIndex

- Needs to update that the Paragraph now has an incoming link from the Document
- The old paragraph also still needs a reference
- It is entirely possible that in this Atom, both paragraphs are modified (and not just one new appended item). Just to make sure, we will first have to delete all references from the old atom, and then create the new references from the new atom. 
- This will be an unnecessarily costly operation if we only append an item to an array

## QueryIndex

- If someone has a Query that sorts document by their number of Paragraphs, their query might need to be updated.
- Note that in this case, the IndexAtom will not have any references - it should have a count


## Solution

I think we need to have two value fields in a QueryAtom: both a Filtervalue and a Sortvalue. Or we should accept that we can't sort by count. Which is also acceptable.

joepio:
I think I finally know what is causing my last failing test: `remove_atom_from_index` and `add_atom_to_index` should get passed a different resource, not the same one. Should update this in the `apply` method of commits

joepio:
Tests are green! I think I'll merge this soon. But first, I'll look for some extra things to break, test and fix.

PR #379, opened 2022-04-11
Failed test db resources all - warm up for threads and async resource fetch

AlexMikhalev:
First attempt to improve all_resources:
1) changed `let subject_bin = subject.as_bytes()` - same encoding (bincode zeropad prefix), that allows to use scan_prefix for sled::Tree which logically fit 
2) Added cache_capacity(1000 * 1024 * 1024) to see if it improves benchmarks, I think it shall be in the config with a threshold value for the percentage of available memory
3) turned for into map if you want all_resources to be iterator we can remove the last collect call. 
Welcome for feedback, I plan to try various multi-thread and async runtimes to see if it improves performance. 
I also think we shall make include_external part of config and global variable.


AlexMikhalev:
Apologies, I forgot to squash commit.

lib/src/db.rs
```diff
@@ -107,7 +112,7 @@ impl Db {
     #[instrument(skip(self))]
     fn set_propvals(&self, subject: &str, propvals: &PropVals) -> AtomicResult<()> {
         let resource_bin = bincode::serialize(propvals)?;
-        let subject_bin = bincode::serialize(subject)?;
+        let subject_bin = subject.as_bytes();
```
joepio, line 115: This may cause DB corruption, as I'm not sure that these serialize in the same way.

If it does corrup between versions, we need some form of (automated) migration.

lib/src/db.rs
```diff
@@ -467,21 +471,26 @@ impl Storelike for Db {
 
     #[instrument(skip(self))]
     fn all_resources(&self, include_external: bool) -> ResourceCollection {
-        let mut resources: ResourceCollection = Vec::new();
-        let self_url = self
-            .get_self_url()
-            .expect("No self URL set, is required in DB");
-        for item in self.resources.into_iter() {
+        // FIXME: Try rossbeam channels here
```
joepio, line 474: What are rossbeam channels? Maybe create an issue for this and mention it here?

lib/src/db.rs
```diff
@@ -628,6 +637,179 @@ impl Storelike for Db {
             }
         }
     }
+
+    fn add_resource(&self, resource: &Resource) -> AtomicResult<()> {
```
joepio, line 641: I'm a bit of confused with the next few hundred LOC. Seems like this is the same as the functions in the `StoreLike` Trait. Why have you copied them? Did it not compile if you kept them in the Trait?

lib/src/db/test.rs
```diff
@@ -385,14 +385,27 @@ fn query_include_external() {
...
-    println!("{:?}", res_no_include.subjects.len());
+    println!("***Query  {:?}", res_include.subjects.len());
+    println!("***Query no include {:?}", res_no_include.subjects.len());
     assert!(
         res_include.subjects.len() > res_no_include.subjects.len(),
         "Amount of results should be higher for include_external"
     );
 }
 
+// check if all_resources include_external working
```
joepio, line 396: Yay tests!

lib/src/storelike.rs
```diff
@@ -63,6 +63,8 @@ pub trait Storelike: Sized {
     /// WARNING: This could be very expensive!
     fn all_resources(&self, include_external: bool) -> ResourceCollection;
 
+    // fn resource_iter(&self) -> Resource;
```
joepio, line 66: I think this should not be here?

joepio (changes requested):
Thank you so much for your PR! Great work :)

I've put some comments in the code. Let me know if I can help out.

lib/Cargo.toml
```diff
@@ -32,6 +32,8 @@ tracing = "0.1"
 ureq = "2"
 url = "2"
 urlencoding = "2"
+futures = "0.3.21"
```
joepio, line 35: I don't see the dependencies being used anywhere yet. Am I overlooking something?

AlexMikhalev: nope, work in progress slipped in.

lib/src/storelike.rs
AlexMikhalev: Correct.

lib/src/db.rs
AlexMikhalev: Yes, it's breaking change and will corrupt the database:
```rust 
let subject="https://atomicdata.dev/datatypes/float";
subject.as_bytes() // new way
[104, 116, 116, 112, 115, 58, 47, 47, 97, 116, 111, 109, 105, 99, 100, 97, 116, 97, 46, 100, 101, 118, 47, 100, 97, 116, 97, 116, 121, 112, 101, 115, 47, 102, 108, 111, 97, 116]
bincode::serialize(subject).unwrap() //old way
[38, 0, 0, 0, 0, 0, 0, 0, 104, 116, 116, 112, 115, 58, 47, 47, 97, 116, 111, 109, 105, 99, 100, 97, 116, 97, 46, 100, 101, 118, 47, 100, 97, 116, 97, 116, 121, 112, 101, 115, 47, 102, 108, 111, 97, 116]
bincode::serialize("https://atomicdata.dev/").unwrap() //mind prefix 
[23, 0, 0, 0, 0, 0, 0, 0, 104, 116, 116, 112, 115, 58, 47, 47, 97, 116, 111, 109, 105, 99, 100, 97, 116, 97, 46, 100, 101, 118, 47]
```
In the old way, serialisation with the same prefix URL will serialize into different "buckets/leafs" - this prevents the use of scan_prefix use or leveraging internal structure for sled::Tree. 
So, what is the approach to data migrations?

lib/src/db.rs
AlexMikhalev: This is my todo.

lib/src/db.rs
AlexMikhalev: That's interesting because I had this question for you - I noticed the same methods in StoreLike and db.rs. Did I grab the wrong branch to play with? I believe db.rs only triggered with --all-features flag.

lib/src/db.rs
joepio: > In the old way, serialisation with the same prefix URL will serialize into different "buckets/leafs" - this prevents the use of scan_prefix use or leveraging internal structure for sled::Tree.

Can't say I fully understand this. `scan_prefix` should also work if there is a shared prefix, in the URL, right? I think I'm missing something here.

> So, what is the approach to data migrations?

I think, sooner or later, we'll need to store the version in a `Tree`, and check this version on booting. If there is a mismatch, the relevant migration script(s) will run. 

So in this case, we iterate over all resources, use `bincode::deserialize` and then put them back with `.as_bytes()`.

We should add a new file in the `db` folder named `migrations.rs`. You then create a function `v0.31_to_v0.32`, which takes the `Db` as its only argument.

When running `Db::init`, we should open a new `Tree` called `Config`, which has a key called `version`, which is checked. If there is no version, we call the new migration.

What do you think?

AlexMikhalev:
As discussed I will create two new from the master branch.

joepio:
https://github.com/joepio/atomic-data-rust/pull/382

@AlexMikhalev i think it may be better to use this instead of master

PR #389, opened 2022-04-19
Memory usage measurement

AlexMikhalev:
This is another PR that is for discussion: 
I would like to propose a tool for measuring consumed memory by data structures, I amended the example
`cargo run --example basic` 
it works on Linux and thanks to Apple botching jmalloc doesn't work on Mac anymore. 
```
Before "593kb"
All resources - include_external: false
96
After "596kb"
All resources - include_external: true
96
```
It uses crate https://github.com/AlexMikhalev/profile which I extracted out of rust-analyzer and Rust analyzer uses memory measurement for a very interesting use case - purging large cache items. See [this](https://github.com/rust-lang/rust-analyzer/blob/b988c6f84e06bdc5562c70f28586b9eeaae3a39c/crates/ide_db/src/apply_change.rs#L104-L238).


joepio:
Great stuff!

Couple of points:

- Rebase onto master, most changes in the PR contain changes that have been merged in already
- Do you plan on maintaining your `profile` repo? Having a simple memory analyzer tool does seem useful, but some already exist: https://crates.io/crates/gperftools, https://github.com/svenstaro/cargo-profiler https://lib.rs/crates/memory-stats
- No mac support is, for me, a tough sell... I use a mac for most of my development. 
- I think this may warrant a feature flag, or only be compiled in dev builds. What do you think?

AlexMikhalev:
I think for now we can add it into dev dependencies - if you can help me on how I can use it inside server lib when developing, maybe via `debug!` trait? I am not sure how current instrumentation works: I saw 'iai' crate in Cargo, but I haven't seen it used. 

Origin of the profile crate lives inside [Rust Analyzer ](https://github.com/rust-lang/rust-analyzer/tree/master/crates/profile) and I hope those fellows will maintain it including mac support, there is a newer version of dependency - arena allocator, I will update once it will be on crates.io. 
All crates you mention rely on external performance tools like valgrind and pain in the back to use, hence I like this crate - it measures heap memory. This PR code was also done on mac, except final commit, I hope rust-analyzer flock will fix it.

AlexMikhalev:
I moved it to dev dependencies and added fix for Mac:
```
Before "35kb"
All resources - include_external: false
96
After "35kb"
All resources - include_external: true
96
```
not sure it works though - the numbers are the same, they are different on linux.

PR #417, opened 2022-05-29
fix: handle removes in commit

rasendubi:
e449e83cb67ca4347627d1e5953145d22e8bfa66 has (accidentally?) removed handling of `remove` field on commits, so removing properties no longer works.

This PR restores it.

Tested it locally with my client code.

PR Checklist:

- [ ] Link to related issue
- [ ] Added tests (if needed)


joepio:
Thanks a ton! I should totally add tests for this... Kind of embarrassing!

PR #419, opened 2022-06-02
fix: fix updating index on commit.remove

rasendubi:
Now it's my time to be embarrassed 😬 

I introduced an error in #417: when `update_index == true`, `resource.get()` would try to get just removed property and would fail, resulting in changes not being saved.

Rework this part, so it gets property from unedited resource. And also handles the case when property does not exist by ignoring it—this works better then erroring out because current clients do not check for property existence before removing.

PR Checklist:

- [ ] Link to related issue
- [ ] Added tests (if needed)

### DIFF
--- a/lib/src/commit.rs
+++ b/lib/src/commit.rs
@@ -254,7 +254,7 @@ impl Commit {
         Ok(commit_response)
     }
 
-    /// Updates the values in the Resource according to the `set`, `remove` and `destroy` attributes in the Commit.
+    /// Updates the values in the Resource according to the `set`, `remove`, `push`, and `destroy` attributes in the Commit.
     /// Optionally also updates the index in the Store.
     /// The Old Resource is only needed when `update_index` is true, and is used for checking
     #[tracing::instrument(skip(store))]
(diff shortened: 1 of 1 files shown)


lib/src/commit.rs
```diff
@@ -274,9 +274,16 @@ impl Commit {
                 resource.remove_propval(prop);
 
                 if update_index {
-                    let val = resource.get(prop)?;
-                    let atom = Atom::new(resource.get_subject().clone(), prop.into(), val.clone());
-                    remove_atoms.push(atom);
+                    if let Ok(val) = resource_unedited.get(prop) {
+                        let atom = Atom::new(resource.get_subject().clone(), prop.into(), val.clone());
```
rasendubi, line 278: As another option, instead of calling `apply_changes()` twice with `update_index` set to false and true, `apply_changes()` might additionally return `remove_atoms` and `add_atoms`, and `apply_opts` would apply them.

joepio: That's a clever idea! :)

joepio (approved):
Seems like a nice fix, thank you so much for taking the time to understand atomic data and to work on this issue!

PR #449, opened 2022-06-15
Try docker dev container

joepio:
Docker can be used to run development containers. This is especially well integrated with VSCode. This could help new developers get their dev setup up and running. No more hassle with `yarn`, `node`, `cargo` versions and all that - all dev tooling is included in the image.

Unfortunately, I can't seem to run `cargo run` in the container, as it runs out of memory during `fetch`:

```
Updating crates.io index
Killed Fetch [======>                  ]  29.46%, 1.29MiB/s    
```

I can see the memory limit getting closer (I provided 11GB) and the app crashes when it is reached.

But `cargo run` should have means to respect memory limits... What's going on here?

Maybe [Hitting Container Kernel Memory Limit causes OOM on Docker Host](https://github.com/docker/for-linux/issues/1001)?

PR Checklist:

- [ ] Link to related issue
- [ ] Added tests (if needed)

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,9 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.238.0/containers/rust/.devcontainer/base.Dockerfile
+
+# [Choice] Debian OS version (use bullseye on local arm64/Apple Silicon): buster, bullseye
+ARG VARIANT="buster"
+FROM mcr.microsoft.com/vscode/devcontainers/rust:0-${VARIANT}
+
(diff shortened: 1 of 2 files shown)


AlexMikhalev:
Do you want me to look into it? Docker + sccache can speed up re-build. We can also put it into github workflows.

joepio:
@AlexMikhalev If you feel like it, sure! I have no clue where to find a solution for this. Note that for me there is no direct reason to make this thing work - it's more just something I wanted to play with.

I hadn't even considered that this could speed up build times, but it makes a ton of sense.

joepio:
I might explore this further, but for now I don't think it has serious advantages over a regular setup.

PR #463, opened 2022-07-05
Fluvio - secret sauce for scalability

joepio:
_By @AlexMikhalev, replaces #461_

## Getting Started - warm up and machine to machine communications

```
minikube status
fluvio profile current
fluvio cluster check
```

Code quality "Hello World", for discussion and not merge - large change and it's combined "Hello World" for Fluvio Rust library and examples/basic.rs
First modify basic example to emit new article on successful save: 
record()
Second add consume_record, where we are:
1) Reading fluvio stream
2) parsing key into subject (key for Atomic Data Server)
3) Parsing value into resource
4) If successful saving into local (in memory in this example storage)

Start with 
1) `cargo run --example async_std_consumer --all-features -- produce_record`
2) Check that data is available in topic `fluvio consume atomic-data -B`
3) `cargo run --example async_std_consumer --all-features -- consume_record`  

That closes Machine 2 Machine communication discussion: https://github.com/atomicdata-dev/atomic-data-rust/issues/425 . Obviously we can now extend commit hook to produce record into fluvio topic in the same way as you added tantivy indexing.

## Now cool stuff
We have: JSON_AD importer
https://github.com/atomicdata-dev/atomic-data-rust/issues/390

Atomizer 
https://github.com/atomicdata-dev/atomic-data-rust/issues/434  

Wasm plugins 
https://github.com/atomicdata-dev/atomic-data-rust/issues/73  

discussions and roadmap items. As it happens Fluvio supports HTTP (and other) connectors, which allow us to run [Smartmodule (WASM plugins) ](https://www.fluvio.io/smartmodules/)on "the edge" - before data arrives into fluvio topic. 
So we can build a highly scalable Atomizer with an option to pre-process data "on-fetch" and "on-save". If it sounds like dream come true, this is because it is ;-) 

Lets start by building fetcher for JSON-AD from reference-architecture.ai
Follow:
https://www.fluvio.io/smartmodules/transform/array-map/ 
I created `array-map-ref-arch-ad` don't use underscores in name of connector or topic - fluvio doesn't like it in many places. 

```bash
cd array-map-ref-arch-ad
cargo build --release

fluvio smart-module list 
(should be empty)
fluvio smart-module create array-map-ref-arch-ad --wasm-file ./target/wasm32-unknown-unknown/release/array_map_ref_arch_ad.wasm
fluvio smart-module list
  NAME                   STATUS             SIZE   
  array-map-ref-arch-ad  SmartModuleStatus  156905
```

Now start http connector locally:
```bash
docker run -d --name="my-http" \
-v"$HOME/.fluvio/config:/home/fluvio/.fluvio/config" \
-t infinyon/fluvio-connect-http:latest \
-- \
--endpoint="https://reference-architecture.ai/json-ad" \
--fluvio-topic="atomic-data" \
--interval=10s --arraymap="array-map-ref-arch-ad"

```
Http connector can be deployed into Infinyon cloud or run locally, since I use minikube it can't fetch externally so it fails. #FIXME

Connector parameters can be specified reference_architecture.yml and registered on cloud with 

`fluvio connector create --config=./reference_architecture.yml`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -421,6 +421,166 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-attributes"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a3203e79f4dd9bdda415ed03cf14dae5a2bf775c683a00f94e9cd1faf0f596e5"
+dependencies = [
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "async-channel"
+version = "1.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2114d64672151c0c5eaa5e131ec84a74f06e1e559830dabba01ca30605d66319"
+dependencies = [
+ "concurrent-queue",
+ "event-listener",
+ "futures-core",
+]
+
+[[package]]
+name = "async-executor"
+version = "1.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "871f9bb5e0a22eeb7e8cf16641feb87c9dc67032ccf8ff49e772eb9941d3a965"
+dependencies = [
+ "async-task",
+ "concurrent-queue",
+ "fastrand",
+ "futures-lite",
+ "once_cell",
+ "slab",
+]
+
+[[package]]
+name = "async-global-executor"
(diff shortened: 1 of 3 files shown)


AlexMikhalev:
Not relevant any more.

Issue #276, opened 2022-01-07
E-mail support


joepio:
At some point, we'll want to send e-mails. For notifications, registrations, or newsletters, for example.

I think it makes sense to offer an e-mail API at the plugin level.

A candidate is [lettre](https://github.com/lettre/lettre).

joepio:
I've managed to get `mail-send` working with a local SMTP server (although not properly embedded in Atomic-Server).

The following approach, using `mailin_embedded`, isn't working for me. The `Transport::new` from `mail-send` seems to take forever, even if I link

```rust
use mailin_embedded::{AuthMechanism, Handler, Server, SslConfig};
use tracing::info;

use crate::errors::AtomicServerResult;

#[derive(Clone)]
struct MailHandler {}
impl Handler for MailHandler {}

const ADDRESS: &str = "127.0.0.1:12041";
const ADDRESS_MAILCATCHER: &str = "127.0.0.1:1025";

pub fn start_mail_server() -> AtomicServerResult<()> {
    let handler = MailHandler {};
    let mut server = Server::new(handler);
    let address = ADDRESS;

    server
        .with_name("example.com")
        .with_auth(AuthMechanism::Plain)
        .with_ssl(SslConfig::None)
        .map_err(|e| format!("Mail SSL error: {e}"))?
        .with_addr(address)
        .map_err(|e| format!("Mail address error: {e}"))?;

    tokio::spawn(async move { send_mail().await });

    tokio::task::spawn_blocking(move || {
        info!("Starting mail server at {address}");
        // TODO: stop this process on ctrl+C
        server.serve().expect("Failed to start mail server");
    });

    Ok(())
}

pub async fn send_mail() -> AtomicServerResult<()> {
    use mail_send::{mail_builder::MessageBuilder, Transport};

    // Build a simple multipart message
    let message = MessageBuilder::new()
        .from(("Joep Meindertsma", "joep@ontola.io"))
        .to(vec![("Joep Floep", "joepmeindertsma@gmail.com")])
        .subject("Hi!")
        .html_body("<h1>Hello, world!</h1>")
        .text_body("Hello world!");

    info!("Sending mail");
    // Connect to an SMTP relay server over TLS and
    // authenticate using the provided credentials.
    Transport::new(ADDRESS_MAILCATCHER)
        .connect()
        .await
        .map_err(|e| e.to_string())?
        .send(message)
        .await
        .map_err(|e| e.to_string())?;
    info!("Successfully sent mail");

[...]

PR #505, opened 2022-09-26
Sudomain support, register endpoint

### LINKED ISSUE #288: Multi-tenancy
Let's consider what it would take to give users / organisations their own store.

## One `Tree` per tenant

We use a bunch of `sled`'s `Tree` items to store stuff on the disk. We do this for Resources and various indexes.

### LINKED ISSUE #489: sign-up endpoint: create a new Drive and a new Agent in one request
Creating a new user / Agent is currently always done using Invites. That flow doesn't make a lot of sense for most usecases: users will probably want their own workspace / Drive, and not join some existing one.

So how should that work? I think we need a `/signUp` endpoint that takes a public key + username and creates a drive + agent.

### LINKED ISSUE #502: Wildcard subdomains
I'm trying to find out if we can host multiple subdomains from actix-web with TLS enabled. See #288.

### LINKED ISSUE #556: Constrain new URLs
At this moment, Atomic Server is very liberal in accepting Commits that create new resources. This can lead to weird situations, that should be prevented:

- Create resources on non-existing drives / subdomains.
- Create resources on drives owned by someone else. (e.g. `john` using the `mary.com/hi` URL)

This should definitely not be allowed.

How can we deal with this?

## Check for `append` right in path

User creates `example.com/topics/1/test/2`, we check the resources `example.com`, `example.com/topics`, `example.com/topics/1` and `example.com/topics/1/test`. If the user has an `append` right on _any_ of these resources, we allow minting the URL. If not, the user cannot create the resource.

- The amount of checks is far higher than before, perhaps making things slower.
- The parent may still be a different, even completely external resource

## Check if the `parent` is in the path of the new URL

User creates `example.com/topics/1/test/2`, we need to check that the parent of that resource is part of that URL, such as `example.com` or `example.com/topics/1`.  

If that parent matches, we can simply perform the regular check at the parent. 

This would be a far more severe limitation, but would help to keep things organised and predictable.

- No additional performance cost, checking if parent matches is trivial
- Users can still create nested resources easily
- We can still have sub-paths that do not exist, e.g. `example.com/topics/1` may not exist, even if `/topics/1/test` does exist. This can be confusing.
- Parent-child relations can no longer span across devices. This is probably a good thing in many situations (because: how do you query foreign machines?), but it also severely limits the flexibility of hierarchies.

joepio:
- closes #502
- closes #288
- closes #489 
- closes #556 
- relates to #276 
- relates to #544 

TODO:

- [ ] Use [new actix `.full_uri()` method](https://github.com/actix/actix-web/issues/2895) instead of some hacky reconstruction. Will be added in [4.8 probably](https://github.com/actix/actix-web/pull/3096#pullrequestreview-2106520805)?
- [ ] Migrate trailing slash base URLs #895
- [ ] Run this on some new domain, check if everything works over there
- [x] Fix double path slash`//` in search results (and other places?)
- [x] Find all the `format!(` occurences where we combine things to create a URL, replace with `URL::join`.
- [ ] Find a strategy to deal with going to a subdomain and losing your local settings (specifically your key!)
- [ ] Consider invite-only registration
- [ ] Fix `all_resources` check for `include_external`. Currently does not properly use subdomains.
- [ ] Add E2E test for registration
- [ ] Remove the existing `new drive` behavior
- [x] Make sure users can't create resources on non-existing drives / subdomains. #556 
- [x] E-mail registration + token support #276 #544
- [x] `/resetPubkey` type of endpoint, in case you lose your private key
- [ ] Fix changelog (currently incorrectly changes previous releases)
- [ ] Tenantized querie #570

PR Checklist:

- [x] Link to related issue 
- [ ] Added tests (if needed)


joepio:

PR #528, opened 2022-10-28
Cookie authentication bug #525

### LINKED ISSUE #525: Cookie authentication not properly working
In the current release, I get a `401` after opening my drive. The error from the server shows that it has not found an Agent, which should imply there is no cookie set. However, there is in fact a cookie sent.

I'm not sure why this isn't working, but I suspect that one of the functions in the cookie parse logic is throwing an error which is thrown away with `.ok()`, which makes it impossible to see what is going on. So even though I'm not sure what is causing the issue, a good first step would be to refactor `get_auth_from_cookie` and make sure it gets rid of the `.ok` calls. 

The reason I think the parsing fails, is because in production, we have multiple cookies. Google analytics is probably responsible for setting the multiple cookies. Our implementation should deal with this.

Screenshot 2022-10-27 at 08 59 15

Using `actix_web::cookie` might be a good idea. 

~~However, even if we update the back-end, we might still have an issue with atomic-data-browser. When it sets the cookies, it may remove others (e.g. google analaytics)~~ <- That is false.

### LINKED ISSUE #527: `Wrong requested subject` when opening atomicdata.dev as agent
- sign in to atomicdata.dev
- refresh the page
- `Wrong requested subject`

This arises from the `get_auth_from_cookie` function. Relates to #525  #520

rescribet:
PR Checklist:

- [x] Link to related issues: closes #525 closes #527
- [ ] Add tests (optional)

### DIFF
--- a/server/src/helpers.rs
+++ b/server/src/helpers.rs
@@ -4,6 +4,7 @@ use actix_web::cookie::Cookie;
 use actix_web::http::header::{HeaderMap, HeaderValue};
 use atomic_lib::authentication::AuthValues;
 use percent_encoding::percent_decode_str;
+use std::str::FromStr;
 
 use crate::errors::{AppErrorType, AtomicServerError};
 use crate::{appstate::AppState, content_types::ContentType, errors::AtomicServerResult};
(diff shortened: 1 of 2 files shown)


server/src/helpers.rs
```diff
@@ -59,42 +61,76 @@ pub fn get_auth_headers(
     }
 }
 
+fn origin(url: &str) -> String {
```
joepio, line 64: does the `actix_web::http:Uri` not provide this with the [`authority` method](https://docs.rs/actix-web/latest/actix_web/http/uri/struct.Uri.html#method.authority)?

Might handle some extra edge cases, I'm not sure.

server/src/helpers.rs
```diff
@@ -59,42 +61,76 @@ pub fn get_auth_headers(
...
+        parsed.authority().unwrap()
+    )
+}
+
 pub fn get_auth_from_cookie(
     map: &HeaderMap,
     requested_subject: &String,
-) -> Option<AtomicServerResult<Option<AuthValues>>> {
-    let encoded_session = session_cookie_from_header(map.get("Cookie")?)?;
+) -> AtomicServerResult<Option<AuthValues>> {
```
joepio, line 77: Looks better, nice!

joepio:
Looks good! I've deployed this version on `atomicdata.dev`, let's hope this also fixes #527

joepio (approved):

server/src/helpers.rs
rescribet: That doesn't include the scheme, so it's not a valid uri (which is the type of the subject field afaict).

Issue #395, opened 2022-04-22
Setting new parent does not correctly update some indexes


joepio:
Locally I noticed that when working on a query with `(some, some, some)` query filter, the index was not updated when I set a new value of a parent. Specifically, the ChatRoom still showed the Message even after I changed the parent of the message to something else. Deletion worked fine, though.

I think `remove_atom_from_index` is calling, but `should_update` is returning false. Not sure about this.

EDIT: 
Here's what's happening. The changed Atom is properly added, and the old one is properly removed. _However_ the change in resource has modified for which Queries it actually fits. So because now the `parent` is different, some QueryFilters should no longer match.

The inverse should probably also be true - let's say the change causes a _new_ QueryFilter to match, that did not match previously. 

I think this can happen if there is `Some(sorted_prop)` and one or two filters (i.e. value, property or both).

So, we only update the Atom for the new `parent`. But in this case, the indexed_atom for the `created_at` is causing the actual problem. That atom is not actually removed, since `created_at` is not mutated.

Solutions:

## Remove all atoms for old resource

Seems pretty simple, but also costly. It means that even if we only want to update one value, we have to remove and then add _all_ indexed items of the resource.

## Change invalidation logic

So let's assume that we only know the updated atom. Is there a way we can find the existing indexed_atom that should no longer match?

We could find all indexed items where the subject matches, and check if the new resource still fits the query. If the answer is no, we remove the atom from the index.

So, if we `check_if_atom_matches_watched_query_filters` for the removal of the old atom, and we get a `resource_match` for prop `p`, we need to check the following:

does the `resource_old` match, and does `resource_new` match? No worries. Does `resource_old` match, but `resource_new` does not? Remove the index atoms!

joepio:
Fixed!

PR #546, opened 2022-12-03
#545 fix indexing logic

### LINKED ISSUE #545: Index logic incorrectly removes items from queries
- Create a collection, with `Some(property)` and `Some(value)` and `Some(sort_prop)`. 
- Create a resource that matches these
- Add irrelevant propval X to this resource
- Remove X
- The collection is now missing X, because there is a bug in the indexing logic

How to fix this problem:

- Add a test for this situation. Either add it to the big `query` test, or create a new one.
- Check the side-effect in `query_index::should_update`, especially [this part.](https://github.com/atomicdata-dev/atomic-data-rust/blob/ee932ca87ec4447b1e2ec4c035beaeaafde6b668/lib/src/db/query_index.rs#L214)

joepio:
PR Checklist:

- [x] Link to related issues: closes #545, relates to #395
- [x] Add tests (optional)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ jobs:
     steps:
       - run: |
           sudo apt-get update
-          sudo apt-get install -y webkit2gtk-4.0 libappindicator3-dev libayatana-appindicator3-dev
+          sudo apt-get install -y webkit2gtk-4.0  libayatana-appindicator3-dev
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
@@ -26,7 +26,7 @@ jobs:
     steps:
       - run: |
           sudo apt-get update
-          sudo apt-get install -y webkit2gtk-4.0 libappindicator3-dev libayatana-appindicator3-dev
+          sudo apt-get install -y webkit2gtk-4.0  libayatana-appindicator3-dev
       - uses: actions/checkout@v2
       - uses: taiki-e/install-action@nextest
       - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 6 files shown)


joepio:
@Polleps I've added a test that demonstrates the issue. You had some ideas on refactoring the indexing logic.

After making the changes, test with `cargo nextest run db::test::index_invalidate_cache`

joepio:
The issue is gone when we remove the whole chunk that was introduced by #395.

The problem was this:

- we have a Query with a filter for `blogpost` items, and a sort by `description`
- we now have an index item containing the description field
- we change the `blogpost` to `recipe`. An IndexAtom (containing `blogpost`) is removed from the indexes.
- the IndexAtom _does not contain the description field, because that wasn't updated_. 
- the IndexAtom created in step 2 still exists!

I don't think I have a test for this, because right now everything seems to pass locally. So I think this PR, as of now, will re-introcuce #395.

The solution was the chunk below in `should_update`

```rust
    if let Some(sort_prop) = &q_filter.sort_by {
        // Sometimes, a removed atom should also invalidate other IndexAtoms, because the QueryFilter no longer matches.
        // This only happens when there is a `sort_by` in the QueryFilter.
        // We then make sure to also update the sort_by value.
        if let Ok(sorted_val) = resource.get(sort_prop) {
            // Note that updating here is a bit too agressive. It will update the index every time an atom comes by that matches the QueryFilter, even if the value is unchanged.
            if &index_atom.property == sort_prop {
                update_indexed_member(
                    store,
                    q_filter,
                    &index_atom.subject,
                    &sorted_val.to_sortable_string(),
                    delete,
                )?;
            }
        }
    }
```

But that solution caused the issue described here, so I removed it now.

We need two things:

- A failing test for #395 
- A fix for that that doesnt introduce new bugs like this one

joepio:
I've just added a failing test for the #395 case. @Polleps feel free to make your changes and run `cargo nextest run db::test::index_invalidate_cache`

joepio:
E2E tests passing locally.

PR #571, opened 2023-02-02
update JS assets

joepio:
Haven't deployed in a while. If this succeeds, I'll deploy.


joepio:
e2e succeeding locally

PR #619, opened 2023-03-13
musl docker #618

### LINKED ISSUE #618: Dockerfile broken `exec /atomic-server-bin: no such file or directory`
I'm pretty sure this issue appeared when I switched to the `rust` base image because of #605, which was because the `cargo` version used didn't recognized the newer `let ... else` syntax. 

Now I get this:

```
exec /atomic-server-bin: no such file or directory
```

Which is weird, because the file is there.

```
/ # ls -lh
total 45M
-rwxr-xr-x    1 root     root       44.6M Mar 14 09:08 atomic-server-bin
```

It's probably a dependency issue in the build. ~~Let's try musl~~ [no doesn't work](https://github.com/atomicdata-dev/atomic-data-rust/issues/618).

I tried using a different base image (`alpine:latest` instead of `frolvlad/alpine-glibc`) but the issue is still there.

Adding a bunch of glibc seemed to help

```
RUN apk add libgcc libc6-compat gcompat
```

But I still get this one:

```
Error relocating /atomic-server-bin: __res_init: symbol not found
```

[Relevant article](https://saarw.github.io/dev/2020/06/18/shipping-linux-binaries-that-dont-break-with-rust.html)

joepio:
Looks like compiling this on mac does not work, but if the buildx works in the CI, i'm keeping it.

PR Checklist:

- [x] Link to related issues: closes #618 
- [ ] Add tests (optional)

Thanks to @AlexMikhalev !

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,5 @@
 target
-Dockerfile
+dockerfile
 .dockerignore
 .env
 .git
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -41,6 +41,8 @@ jobs:
           images: joepmeneer/atomic-server
           github-token: ${{ secrets.GITHUB_TOKEN }}
           tags: |
+            # set latest tag for master branch
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'master') }}
             type=schedule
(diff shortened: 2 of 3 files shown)


joepio:
We already use MUSL with earthly

PR #621, opened 2023-03-14
Add tests for docker image

### LINKED ISSUE #606: CI: test docker build before deploying
We do end-to-end tests in the CI, but the dockerfile generation is actually performed as a _separate_ action. 

What can we do to improve this?

- Run the `docker` job after all other steps from `main` using the `needs` keyword. However, this still allows issues with the docker image itself (e.g. missing a port forward, or some compatibility issue with the image). Also, it might run too often this way.
- In the `docker` job, run the `e2e` tests after building the image, but before pushing it. This requires some changes, because currently one action does both the pushing and the building.

### LINKED ISSUE #618: Dockerfile broken `exec /atomic-server-bin: no such file or directory`

joepio:
Todo:

- [ ] Make sure the docker container runs and the tests succeed!  Currently it does not [recognize the tag](https://github.com/atomicdata-dev/atomic-data-rust/actions/runs/4415251819/jobs/7737978657#step:8:19).

PR Checklist:

- [x] Link to related issues: closes #606
- [ ] Add tests (optional)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -55,6 +55,7 @@ jobs:
           context: .
           push: false
           provenance: false
+          load: true
           tags: joepmeneer/atomic-server:local
 
       - name: Run Docker image
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -53,8 +53,28 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           context: .
+          push: false
+          provenance: false
+          platforms: linux/amd64,linux/arm64
+          tags: local
+          labels: ${{ steps.meta.outputs.labels }}
+
+      - name: Run docker image
+        run: nohup docker run -p 80:80 joepmeneer/atomic-server:local --initialize &
+
+      - name: Setup and run end-to-end tests
+        working-directory: ./server/e2e_tests/
+        run: |
+          pnpm install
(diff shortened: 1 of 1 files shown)


joepio:
This branch is outdated - it uses github instead of earthly. Additionally, earthly does test the executable in the image.

Issue #543, opened 2022-11-24
Feature: Atomic Server can store blobs on S3 compatible storage - like AWS S3, Cloudflare durable objects


AlexMikhalev:
To maintain lightweight for Atomic Server give user the opportunity to add S3 credenitals and endpoint with option to synchronise binary data with external S3 compatible servers. 
I would check how  https://github.com/quickwit-oss/quickwit does it.

joepio:
This would be pretty simple to implement, I think. The logic in `download.rs` and `upload.rs` could be altered depending on whether some `s3` opts (like keys and bucket ID) are being set.

AlexMikhalev:
Yes, and QuickWit have example of downloading from S3 in multithreaded way.

AlexMikhalev:
Prefered way to progress to use OpenDAL https://opendal.apache.org/docs/rust/opendal/services/struct.S3.html

PR #659, opened 2023-09-05
s3 uploads #543

metame:
PR Checklist:

- [x] Link to related issues: #543
- [-] Add tests (optional)
- [x] (If new feature) add to description / readme

Open issues:
1. The data browser preview is currently not working for s3 files. This is likely because the download endpoint does a 302 temporary redirect to a S3 signed URL. I'd be happy to open a new issue for this for someone to fix that is familiar with the client side code.
2. Currently the config only requires S3 bucket to be specified. This is in line with opendal's requirements. The main callout here is that we are depending on opendal's default for region, which is "us-east-1". I could specify this directly in atomic's source code or we can make this a required config value.


joepio:
Awesome, great job! I'll take a closer look tomorrow :)

server/src/files.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+pub async fn s3_upload_object(
+    appstate: &AppState,
+    file_id: &str,
+    file_path: &PathBuf,
+) -> AtomicServerResult<()> {
+    let mut builder = S3::default();
+    builder.bucket(appstate.config.opts.s3_bucket.as_ref().unwrap().as_str());
```
joepio, line 57: Unwrap can't go here, it would cause a crash at runtime

server/src/files.rs
```diff
@@ -0,0 +1,137 @@
+use std::{fmt, path::PathBuf, time::Duration};
+
+use opendal::{services::S3, Operator};
+use tokio::fs::File;
+
+use crate::{appstate::AppState, config::Config, errors::AtomicServerResult};
+
+pub enum FileStore {
+    S3,
+    FS,
```
joepio, line 10: Perhaps this would a good place to store config data? We can then guarantee that if instantiation of the FileStore has succeeded, we've got all the data we need.

We could have a `FileStore::init_from_config` that reads all the relevant props, and then stores the config in the enum.

Not entirely sure if that is the best way though, curious to hear your thoughts.

server/src/files.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+pub async fn s3_upload_object(
+    appstate: &AppState,
+    file_id: &str,
+    file_path: &PathBuf,
+) -> AtomicServerResult<()> {
+    let mut builder = S3::default();
+    builder.bucket(appstate.config.opts.s3_bucket.as_ref().unwrap().as_str());
+    appstate
```
joepio, line 58: We could get rid of these lenthgy calls if the config props are stored in the `FileStore` enum

server/src/handlers/upload.rs
```diff
@@ -50,7 +60,7 @@ pub async fn upload_handler(
 
         std::fs::create_dir_all(&appstate.config.uploads_path)?;
```
joepio, line 61: I think most of the logic here should be moved to the `FileStore` impl. For example, we're calling `create_dir_all` even if the user is using S3, right? And we're creating the `fs_file_id`, whereas we might not need it for S3 uploads.

joepio (changes requested):
Thanks for the PR! Looking good mostly :) 

left some comments, curious to hear your thougts.

server/src/files.rs
metame: Ah yes. I originally implemented this way because we check that the bucket exists before calling this function so I knew it'd exist here, but it definitely introduces some brittleness, especially since it's in its own module. Great call out.

server/src/files.rs
metame: I really like this idea!
I can see two options

1. We add the static `init_from_config` method to the FileStore enum impl, store the config in a private struct, and access the config via a new public method on the enum.  We could access the config struct directly but doesn't feel quite right to me (but open to discussion)
2. We create a new FileStorage struct to store the config, `FileStore`, etc. This initially feels better rather than expanding the enum. We could probably call the struct `FileStore` and the enum `Store` as well. 

Open to discussion on implementation on naming here.

server/src/files.rs
metame: That would make me happy

server/src/handlers/upload.rs
metame: That makes sense to me.

server/src/files.rs
joepio: Not sure if I completely understand storing it in a private struct. It would still be part of the `enum FileStore` than, right?

I'm thinking something like this:

```rust
pub enum FileStore {
   S3(S3Config),
   FS(FSConfig),
}

struct S3Config {
 bucket: String,
 region: String,
  //...etc
}
```

metame: Yes, I'm struggling with thinking through the ergonomics of accessing that config on `FileStore` but I probably just need to code it out.

metame: So it turns out I was just unfamiliar with the Rust Enum tuple-like members. This all makes sense now. :)

server/src/handlers/upload.rs
metame: In this case actually, we are creating an intermediary file on the file system. Typically, if I knew this was a unix host, I would just write the tmp file into `/tmp`, but since we don't know the OS of the host I feel like this makes the most sense at this time. It also provides a fallback in the sense that if a user uploads a file and the s3 upload fails, the file still exists on the server. If the s3 upload succeeds, the file is deleted from the file system. Of course, there are pros and cons to this.

metame:
@joepio This should be ready for another look.

server/src/handlers/upload.rs
metame: I believe I could instead write directly to s3 in the multipart upload handler instead of using the intermediary file. My original thought is this might be more brittle when working with large files vs the current implementation where we essentially "copy" the file from fs to s3. But I will also add that while I've worked with S3 a lot in production code, I have never just "proxied" s3 uploads, typically just used a signed upload URL sent to client or run different processing on a file before uploading to s3.  So my hesitation to directly upload to s3 could just be unfamiliarity on my part.

server/src/config.rs
```diff
@@ -75,6 +75,22 @@ pub struct Opts {
...
+    #[clap(long, env = "ATOMIC_S3_REGION")]
+    pub s3_region: Option<String>,
+
+    /// endpoint for s3-compatible storage service, defaults to "https://s3.amazonaws.com"
+    #[clap(long, env = "ATOMIC_S3_ENDPOINT")]
+    pub s3_endpoint: Option<String>,
+
+    /// path where s3 uploads will be stored
+    #[clap(long, env = "ATOMIC_S3_PATH")]
+    pub s3_path: Option<String>,
```
joepio, line 92: Perhaps also add a token / authentication param. Currently it only works with public buckets, right?

joepio (changes requested):
Looking good! One more general point maybe: Could you add some documentation and tests? I'd accept a PR without them if you had enough of this. 

Especially one happy flow for FileStore in tests would be nice.

server/src/config.rs
metame: There are multiple ways to authenticate. Currently it accepts a credentials file stored at `~/.aws/credentials` just like the aws cli (I'm assuming there's a similar path on windows) or the standard env vars of `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`

metame: But I'm also happy to add the auth options that opendal provides explicitly. We can add the access_key/secret_access_key and the token auth options

server/src/files.rs
```diff
@@ -0,0 +1,137 @@
...
+                region: opts.s3_region.clone(),
+                path: opts.s3_path.clone().unwrap_or("uploads".to_string()),
+            };
+            FileStore::S3(config)
+        } else {
+            fs_file_store
+        }
+    }
+
+    pub fn get_subject_file_store<'a>(appstate: &'a AppState, subject: &str) -> &'a FileStore {
```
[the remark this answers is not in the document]
metame: They are in this case! I originally implemented it without them and the compiler said they were necessary since I was returning a reference.

server/src/config.rs
metame: My hesitation in adding support for these explicitly is it does complicate the code some, as using key/secret auth is mutually exclusive of using token auth. But of course something that we can handle if you think it has value over and above the authentication methods above.

server/src/files.rs
```diff
@@ -36,9 +36,9 @@ impl FileStore {
 
     pub fn init_from_config(config: &Config, fs_file_store: FileStore) -> FileStore {
         let opts = &config.opts;
-        if opts.s3_bucket.is_some() {
+        if let Some(bucket) = &opts.s3_bucket {
```
joepio, line 39: awwwyeah

server/src/handlers/upload.rs
```diff
@@ -132,6 +109,33 @@ pub async fn upload_handler(
     )?))
 }
 
+
+async fn fs_upload(
```
joepio, line 113: Maybe cleaner to have these as Methods on FileStore? Not sure if you agree. Can imagine it's more discoverable that way.

metame: Yes, still in progress. I just push a lot 🙂

metame:
> Looking good! One more general point maybe: Could you add some documentation and tests? I'd accept a PR without them if you had enough of this.
> 
> Especially one happy flow for FileStore in tests would be nice.

Just tests and documentation left. Should be able to finish this up on Friday.

metame:
Alright @joepio this is ready for another look. Tests are passing and functionality tested with small and large files.

joepio:
Great job! Only thing I still have left is [authentication](https://github.com/atomicdata-dev/atomic-server/pull/659#discussion_r1329660385).

server/src/config.rs
metame: > There are multiple ways to authenticate. Currently it accepts a credentials file stored at `~/.aws/credentials` just like the aws cli (I'm assuming there's a similar path on windows) or the standard env vars of `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`

Wanted to make sure you saw this. This is also talked about in the documentation I wrote. Do you feel this is insufficient? For example, in our production applications at work this is how we authenticate to the AWS sdk.

metame: More details on credential files [here](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-format) which we could add to the documentation. Which is to say, the bucket doesn't have to be public.

metame: If given all of this you would still like an explicit way to set authentication in the code, I certainly can do it.

AlexMikhalev:
You can init via both map and hashmap, passing region and endpoint:
```
fn init_operator_via_map() -> Result<Operator> {
    // setting up the credentials
    let access_key_id =
        env::var("AWS_ACCESS_KEY_ID").expect("AWS_ACCESS_KEY_ID is set and a valid String");
    let secret_access_key =
        env::var("AWS_SECRET_ACCESS_KEY").expect("AWS_ACCESS_KEY_ID is set and a valid String");

    let mut map = HashMap::default();
    map.insert("bucket".to_string(), "test".to_string());
    map.insert("region".to_string(), "us-east-1".to_string());
    map.insert("endpoint".to_string(), "http://rpi4node3:8333".to_string());
    map.insert("access_key_id".to_string(), access_key_id);
    map.insert("secret_access_key".to_string(), secret_access_key);

    let op = Operator::via_map(Scheme::S3, map)?;
    Ok(op)
}
```
This would make Opendal based S3 opened to wider use - minio, cloudflare d2 etc.

AlexMikhalev:
Late to the party, but personally, I would prefer to remove this parameter:
``` 
    /// CAUTION: Skip authentication checks, making all data publicly readable. Improves performance.
    #[clap(long, env = "ATOMIC_PUBLIC_MODE")]
    pub public_mode: bool,
```
It's the job of the atomic server to decide if the file or any other resource is public or visible to the group. Storage is always private with a key/secret pair; if people want to share their s3 bucket as a public website, they have another way of doing it.

joepio:
> Late to the party, but personally, I would prefer to remove this parameter:
> 
> ```
>     /// CAUTION: Skip authentication checks, making all data publicly readable. Improves performance.
>     #[clap(long, env = "ATOMIC_PUBLIC_MODE")]
>     pub public_mode: bool,
> ```
> 
> It's the job of the atomic server to decide if the file or any other resource is public or visible to the group. Storage is always private with a key/secret pair; if people want to share their s3 bucket as a public website, they have another way of doing it.

This flag is unrelated to this PR / S3, it's for regular authorization checks.

joepio:
I think my 404 issue has to do with the URL encoding. The first part in `s3:fileID` is URL encoded, so it becomes `s3%3AfileID` (`:` becomes `%3A`). When we refresh the page, though, the server un-encodes it I think so we get the `:` instead, which 404s. 

@metame do you remember what the encoding was needed for? Sorry I only find out about this so long after you opened this PR!

metame:
> I think my 404 issue has to do with the URL encoding. The first part in `s3:fileID` is URL encoded, so it becomes `s3%3AfileID` (`:` becomes `%3A`). When we refresh the page, though, the server un-encodes it I think so we get the `:` instead, which 404s.
> 
> @metame do you remember what the encoding was needed for? Sorry I only find out about this so long after you opened this PR!

So the main reason we implemented the file prefix was to support backwards compatibility among file systems. That is, if an atomic server instance had files uploaded to the server directly and then turned on s3 support, the existing files would still work.

I've found the source of this bug, but the fix has some drawbacks. Specifically, it requires url-encoding the subject path for every resource that goes through `handler::get_resource` like [here](https://github.com/atomicdata-dev/atomic-server/pull/659/commits/0dde795cd0e94014d4dd1e5e6da645c07e23b499#diff-80b6db7e3eaa6ea54e14747f2c0904dab689dc1d7de90936729a06d43131be71R73). Perhaps, this is acceptable. If not, we could drop the behavior that supports backwards compatibility and server admins would have to deal with the migration themselves.

joepio:

PR #671, opened 2023-10-02
Earthly ci

AlexMikhalev:
Example of using Earthly to build atomic server, 
I don't understand where pnpm shall output folder, ideally you want to run frontend build separately and cache dependency. 
Hack using bun didn't work either as build.rs can't find bun or pnpm.


joepio:
Thanks again! Replaced by #672, which cherry picked the earthly changes + some updates

Issue #799, opened 2024-01-11
Better datatype picker


Polleps:
In normal forms the datatype picker is a `SearchBox`, This means that the user has to know what datatypes exist before they pick one. This wasn't always the case as before the SearchBox it was the old ResourceSelector with all dataypes hardcoded in.
The Ontology editor uses a simple dropdown that is much more intuitive for datatypes than a SearchBox. We should use this component for normal forms too.

Issue #788, opened 2024-01-08
formatToParts error in read mode ontology editor


joepio:
- Create new `Allows-only` in ontology editor
- Don't fill in name
- Go to `read` mode

err:

```
formatter.formatToParts(subjects).map(({ type, value })
```

PR #808, opened 2024-01-17
#780 enums in ontologies

### LINKED ISSUE #780: Instance creation is odd - creates a class
In the ontology editor, creating an instance doesn't quite make sense yet. Instead of creating (almost empty) resources, we create "Class" instances. 

Also, selecting an instance in "allows only" should offer quick creation of instances. (ideally without a form at all)

Polleps:
PR Checklist:

- [x] Link to related issues: closes #780, #799, #788
- [x] Add tests (optional)

### DIFF
--- a/browser/data-browser/src/components/Tag/CreateTagRow.tsx
+++ b/browser/data-browser/src/components/Tag/CreateTagRow.tsx
@@ -2,7 +2,6 @@ import { Resource, core, dataBrowser, useStore } from '@tomic/react';
 import { useState, useCallback, Suspense, lazy } from 'react';
 import { FaPlus } from 'react-icons/fa';
 import { randomItem } from '../../helpers/randomItem';
-import { randomString } from '../../helpers/randomString';
 import { stringToSlug } from '../../helpers/stringToSlug';
 import { Button } from '../Button';
 import { Row } from '../Row';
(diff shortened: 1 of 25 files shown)


browser/data-browser/src/components/Tag/CreateTagRow.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export function CreateTagRow({ parent, onNewTag }: CreateTagRowProps) {
+  const store = useStore();
+  const [tagName, setTagName] = useState<string>('');
+  const [emoji, setEmoji] = useState<string | undefined>();
+  const [resetKey, setResetKey] = useState<number>(0);
+
+  const createNewTag = useCallback(async () => {
+    const tag = await store.newResource({
+      subject: `${parent}/${randomString()}`,
```
joepio, line 27: should use tagname

joepio: Make sure error is catched if name collission

joepio: There is a function that does check this... perhaps use that?

joepio (approved):

joepio (approved):

PR #818, opened 2024-01-23
 #764 Add option to format numbers as currency in tables

### LINKED ISSUE #764: Introduce the financial format for Number Column
image
Currently, only numbers and percentages are available, which makes financial documentation inconvenient. It should be available in a financial format like in Excel, with "," for separating thousands and "()" for negative numbers.

Polleps:
PR Checklist:

- [x] Link to related issues: closes #764 
- [ ] Add tests (optional)


browser/data-browser/src/views/TablePage/helpers/formatNumber.ts
```diff
@@ -21,6 +22,22 @@ export function formatNumber(
...
+    try {
+      const formatter = new Intl.NumberFormat('default', {
+        style: 'currency',
+        currency,
+        currencyDisplay: 'narrowSymbol',
+      });
+
+      return formatter.format(value);
+    } catch (e) {
+      console.error(e);
```
joepio, line 35: This should be `handleError`

joepio (approved):

browser/data-browser/src/views/TablePage/helpers/formatNumber.ts
Polleps: Oh this catch should actually do nothing and just let the formatter continue to format the value as if it were just a number as this only triggers when the currency is not supported by the browser.

PR #858, opened 2024-03-07
 #757 Make sidebar items drag & droppable

### LINKED ISSUE #757: Drag & Drop sidebar - Missing feature of re-arranging the resources
If I am not happy with the sequence, I created the resources:
image
I still cannot re-arrange them in a more logical order. Thus, there should be a way to move them up-down or group them into folders (now I understand only how to create atoms based on types).

Polleps:
PR Checklist:

- [x] Link to related issues: closes #757 
- [ ] Add tests (optional)


joepio:
Great job! Awesome feature.

I tried creating an E2E test for this, but it looks like there's not drag / drop in Playwright (test creator), so I suppose we should just let it exist without a test. I can live with this.

PR #892, opened 2024-06-11
#855 api guide code gen

### LINKED ISSUE #855: Make it easy to use the API from table view
- User has table containing instances
- User wants to fetch the data... somewhere

How do we make their UX as easy as possible?

## Copy-Paste `/query` URL

- Add an action in the Table view for copying the whole `/query` URL, which should return current resources.

## Generate some JS / TS code blocks

- Using `@tomic/lib` stuff.

### LINKED ISSUE #888: Font size not scaling properly
The font size on body is set to 15px. This means that when the browser is configured to display larger text some text in the databrowser will not scale accordingly.

This is obviously bad for accessibility.

The fix is rather simple, just set the bodies font size to `0.95rem`, this way we keep the slightly smaller than default text size but it will scale up if the user needs it.

Polleps:
## Related Issues

closes #855, closes #888

## Checklist
- [ ] Update docs if needed


joepio (approved):

joepio:
very cool!

PR #897, opened 2024-06-17
Clear classtype when setting datatype, if datatype is not resource or resource array

lippinj:
Classtype is only a valid piece of state when datatype is resource or resource array. Setting the datatype to something else now clears the classtype, like it has been clearing the allows-only list.

Also changed `urls` -> `core` because deprecated.

### DIFF
--- a/browser/data-browser/src/views/OntologyPage/PropertyDatatypePicker.tsx
+++ b/browser/data-browser/src/views/OntologyPage/PropertyDatatypePicker.tsx
@@ -1,4 +1,11 @@
-import { Resource, reverseDatatypeMapping, urls, useArray } from '@tomic/react';
+import {
+  Datatype,
+  Resource,
+  reverseDatatypeMapping,
+  core,
+  useArray,
+  useString,
+} from '@tomic/react';
 import { AtomicSelectInput } from '../../components/forms/AtomicSelectInput';
 interface PropertyDatatypePickerProps {
   resource: Resource;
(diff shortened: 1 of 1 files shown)


browser/data-browser/src/views/OntologyPage/PropertyDatatypePicker.tsx
```diff
@@ -16,29 +23,34 @@ export function PropertyDatatypePicker({
...
-      return;
-    }
+  const isResourceLike = (datatype: string) => {
+    return (
+      datatype === Datatype.ATOMIC_URL || datatype === Datatype.RESOURCEARRAY
+    );
+  };
 
-    setAllowsOnly(undefined);
+  const handleChange = (datatype: string) => {
```
Polleps, line 39: The name `handleChange` is a bit too generic. It doesn't actually handle setting the value that changes as that is done inside the `AtomicSelectInput`. Calling it something like `removeTypeSpecificProps` would be better in this case

lippinj: `clearInapplicableProps()`?

lippinj: @Polleps changed the name -- better now?

Polleps: Yes perfect!

PR #959, opened 2024-09-02
Create template #952

### LINKED ISSUE #763: The sequence of properties in card view does not match the one in the table view.
image
image
Solution - when the user changes the column sequence in the table view, it should change the layout in the card view as well.

### LINKED ISSUE #894: What to do about nested dialogs
There are a few situations where you can open a dialog from another dialog. In general we try to avoid these scenarios but they do exist. For example when creating a resource inside you can create nested resources by selecting the 'Create: <name>' option in a resource selector. A dialog will open where you can create the nested resource. If this resource class happens to have a file property, A file selector button will be rendered that opens another dialog.

Nested dialogs are problematic because they are confusing.

There is also a bug where both dialogs are closed when the user closes one of them.

### LINKED ISSUE #952: `create-template` package to `npm create @tomic/template`
We could make a `create-template` package that allows users to do things like `npm create @tomic/template`
Since these are just executable scripts we can do things like generate a users ontologies and have some options like what framework to use.

_Originally posted by @Polleps in https://github.com/atomicdata-dev/atomic-server/issues/923#issuecomment-2304394809_

## Goals

- Make it easier for devs to start a website using `@tomic/lib` and `@tomic/svelte`
- Dogfeed our APIs so we make them as easy as possible for our main usecase (Headless CMS)
- Test libs in a project in the CI (instead of just relying on unit tests)

## TODO

- [x] Template (sveltekit?) project, should be minimal
  - [ ] Render Articles
  - [ ] Render a Collection (using some custom view)
  - [x] Search bar
  - [ ] Readme that mentions the CLI and links to docs where needed
- [ ] Template ontology (article, menu item)
- [ ] cli action that copies a folder
- [ ] link to an ontology to create types
- [ ] e2e tests (should use `pnpm link`)
- [ ] Publish new NPM package
- [ ] Publish it to netlify (and add docs)
- [ ] Publish it to cloudflar (and add docs)

### LINKED ISSUE #955: Save queueing seems broken
When creating a child resource of a new resource, an error occurs when trying to save the child resource.
The error comes from the server and it says it's parent doesn't exist. This is true because we havent saved the parent yet but we have some logic in `Store` that should prevent this and queue the child to save after its parent has finished saving.

### LINKED ISSUE #967: Creating a resource via a resource selector always sets shortname
We set the shortname of the new resource based on the name that was entered in the input:
```ts
initialProps={{
  [urls.properties.shortname]: initialNewTitle,
}}
```

This should be `core.properties.name` for classes that have a name prop instead of shortname.

Also for classes that don't have any kind of title we should maybe just show a `create <class>` option without the users input text

### LINKED ISSUE #968: There is no way to select existing files by subject from the new file picker dialog.
I wanted to select a file from a different drive/server but realised this is not possible now.

Maybe we can make it so the user can paste a subject in the search field to select the resource by subject?

### LINKED ISSUE #969: Markdown editor value is not always saved when saving the resource
Looks like the markdown editor only creates a commit the first time it's loaded.
Probably a memoization issue.

### LINKED ISSUE #975: Create new resources from resources selectors without classtype
Users are able to create new resources from the resource selector but only if the property has a classtype set.
When a property does not have a classtype there won't be a 'create ___' option in the dropdown.

This is confusing and breaks the flow of creating children for new resources.

We need a create option in these resource selectors that let the user pick a class first and then resume the normal resource creation flow.

Polleps:
## Related Issues

closes #952, closes #968, closes #969, closes #967, closes #955, closes #894, closes #975, closes #763

## Checklist
- [x] Update docs if needed


joepio (changes requested):

joepio (approved):
great job!

PR #987, opened 2024-10-09
Fix importer prop sorting causes parents to be parsed after their chi…

### LINKED ISSUE #980: Importing template not working - order
`store.import` takes a JSON-AD string that can have multiple`local-id` items. The current import / parse implementation is _order dependent_, because sometimes a resource can refer to another resource that is not created yet:

- If a `property` is used before it's defined
- If a `parent` is used before it's defined

@Polleps implemented sorting in the parser to prioritize properties, which solves some cases, but there still are failure cases, e.g. when importing the new template (#959).

Polleps:
…ld properties #980

## Related Issues

closes #980

## Checklist
- [ ] Update docs if needed

### DIFF
--- a/lib/src/parse.rs
+++ b/lib/src/parse.rs
@@ -108,6 +108,58 @@ fn object_is_property(object: &serde_json::Value) -> bool {
     false
 }
 
+fn get_parent_to_pull(value: &serde_json::Value) -> Option<String> {
+    if let serde_json::Value::Object(map) = value {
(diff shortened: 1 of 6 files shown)


lib/src/parse.rs
```diff
@@ -127,6 +178,8 @@ pub fn parse_json_ad_string(
                 b_is_prop.cmp(&a_is_prop)
             });
 
+            arr = pull_parents_of_props_to_front(&arr);
```
joepio, line 184: This kinda stuff needs some documentation / comments / explanation IMO

PR #1016, opened 2024-11-20
For discussion: Web components generation

AlexMikhalev:
This PR is for discussion and not merging:
The idea is that it's quite easy to generate web components from atomic data classes by extending the ontology cli command—maybe via an additional flag and saving each web component as a separate file. 
The user can use it:
```
<!-- Example usage -->
<atomic-bookmark 
  subject="https://atomicdata.dev/bookmarks/1">
</atomic-bookmark>
```
I used data browser as ontology.

### DIFF
--- a/browser/cli/atomic.config.json
+++ b/browser/cli/atomic.config.json
@@ -1,0 +1,7 @@
+{
+  "outputFolder": "./src/ontologies",
+  "moduleAlias": "@tomic/lib",
+  "ontologies": [
+    "https://atomicdata.dev/ontology/data-browser"
+  ]
+}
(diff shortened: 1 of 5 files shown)


joepio:
I like this idea! Really interesting workflow.

@RoelLeijser could be relevant to your project.

Polleps:
Interesting Idea but I don't see why you would create a component for each class. You could make a form component that generates the form based on the resources class at runtime. We would also need a way to access the store inside the component. I believe attributes can only hold serializable values (so no class instances).
If there is just one form component there would also be no need to generate it with cli as we could just export it from @tomic/lib

AlexMikhalev:
Form components can be a useful addition to @tomic/lib regardless of the proposed solution, but storing pre-built web components locally allows us to modify styles and UI presentation. At the same time, to maintain consistency between data model and interface we can leverage git to ensure versions of UI and data model/typescript types match.

AlexMikhalev:
@Polleps, that's not far from generating templates for other frameworks. IMHO, the right way to use an atomic server isn't to start with code but with creating a domain ontology. Then, you can generate typescript bindings and corresponding web components, and you can add a flag to generate components for Astro/React/Svelte.

joepio:

Issue #985, opened 2024-10-09
Improve the `basic.rs` client example


joepio:
@AlexMikhalev wanted to have some more simple examples of how to use the `atomic_lib` rust library. The current `basic.rs` misses some features:

## Basics

- [ ] Creating a resource
- [ ] Signing in with agent, getting hidden resources.
- [ ] #1040

## Bonus: 

- [ ] Full text search
- [ ] Collection / pagination

AlexMikhalev:
I have the following flow in mind:
1) Demonstrate how to create a store passing server URL
2) Demonstrate how to create a store with an authenticated agent passing URL an agent string
3) Demonstrate how to read from the store 
4) demonstrate how to write to store
5) Demonstrate how to fetch data from the atomic server using collections 
6) Demonstrate how to search over atomic server using search API (tantivy)
7) add paginations, but I am more interested in retrieving all data and re-ranking them - Terraphim will handle re-ranking.

PR #1021, opened 2024-11-23
Tiny example of how to use query API

AlexMikhalev:
Tiny example of how to use query API

improves on #985


Atishayjain26 (approved):

joepio:
Thank you!

PR #1089, opened 2025-07-01
Ontology nav fix #1083

### LINKED ISSUE #1083: Navigation in the ontology view no longer works
Clicking on one of the links in the sidebar in the ontology editor should move the page down to that class or property but now it tries to load a resource that doesn't exist.

The way these links work is by having `#http://my-subject` at the end of the href so the page scrolls down to the correct anchor tag. Now it tries to fetch the resource including the anchor part at the end which obviously does not work.

Polleps:
## Related Issues

closes #1083 

## Checklist
- [ ] Update docs if needed


joepio:
lgtm

PR #1090, opened 2025-07-01
Fix subresources not updated when resource is deleted #1087

### LINKED ISSUE #1087: Deleting a resource no longer removes it from the sidebar
Deleting a resource no longer removes it from its parents subresource list. I believe this is handled by an event registered on the store, maybe that is no longer triggering.

Polleps:
## Related Issues

closes #1087

## Checklist
- [ ] Update docs if needed


joepio:
lgtm

PR #1103, opened 2025-07-24
Uri and json fields #1099

### LINKED ISSUE #658: JSON datatype
Sometimes you don't want to deal with creating Properties and hosting these somewhere - you just want to have some JSON thing (array, object, whatever). 

We could add JSON as a datatype.

## Implementation

- [ ] Add it to `datatypes.rs`
- [ ] Add to `urls.rs`
- [ ] Add to `values.rs` + test
- [ ] Add to `serialize.rs` (`val_to_serde`)
- [ ] Add to `default_store.json`

### LINKED ISSUE #913: Slow build.rs / rust analyzer
Running `build.rs` has some steps that are sometimes unnecessarily slow. This is particularly annoying because rust-analyzer seems to be slowed down due to this.

### LINKED ISSUE #1024: Support URI datatype
We have a [URI datatype in the docs](https://atomicdata.dev/datatypes/uri), but it's not in the implementations.

## TODO

- [x] rust Value + Datatype
- [x] js Value + Datatype
- [x] browser ontology editor
- [x] browser validation in form
- [ ] browser table editor
- [x] browser preview (render as `a href`)

### LINKED ISSUE #1099: JSON Schema & Atomic Schema
These two specs have some interesting similarities and differences. Let's explore them both and think about potential compatibility. I want to either move Atomic Schema towards JSON Schema, or make some small changes to JSON Schema, or have good reasons to keep both separate.

## Comparison

Both JSON Schema and [Atomic Schema](https://docs.atomicdata.dev/schema/intro.html):

- Make JSON type-safe
- Support linking to schema definitions (in JSON Schema you can share one schema using `$ref`, in Atomic Schema using [`isA`](https://atomicdata.dev/properties/isA)
- Are open
- Have JS + Rust implementations
- Are self-describing (JSON Schema, [Atomic](https://atomicdata.dev/classes/Class))

Only JSON Schema:

- Supports complicated JSON objects, like arrays of certain types of objects
- Has far more other implementations (python, go etc)

Only Atomic Schema:

- Supports de-referencing properties (all keys are HTTP URLs). This is probably one of the biggest differences, and quite fundamental to Atomic Data. The idea is that it is useful to _re-use properties_ in multiple classes. As far as I know this is not a thing in JSON Schema
- Has clear expectations about the protocol for fetching and browsing the data. Every key needs to resolve with an `application/ad+json` request, it has expectations about authentication, etc.

### Generate JSON Schema from Atomic Schema

This would allow other languages to easily validate their Atomic Data client-side, even if there is no library available.

- Add `/endpoints/json-schema` which takes a class, property or ontology and generates a JSON schema that resolves.

### Get rid of Atomic Schema altogether, depend on JSON Schema

- Could we still retrieve JSON Schema definitions for properties and use these to render form fields? If that is the case, maybe this is a legit path forward.

### Allow JSON datatype + JSON Schema property

- Add an implementation that downloads the referenced schema (could happen recursively)

### LINKED ISSUE #1101: Consider alternative to bincode for persisting structs in KV store
Bincode is fast and compact, but it's very sensitive to schema changes. If our `Value` / `DataType` enum changes, the schema might break, which means we need to build a sometimes quite complex migration.

## Messagepack

- Self describing, which means no external schema needed
- A bit less fast and efficient than bincode as it's self-describing.

## Protobuf

- Requires an external schema, which we could probably build in some folder using `build.rs`

### LINKED ISSUE #1107: Remove Named Nested Resources from the JSON-AD spec
Currently it is allowed to have named resources (with an `@id`) as values of properties.
This is used in API's like search, collections and chatrooms to include results in the json response to minimize network roundtrips.

Values of properties are not trusted sources so they should not be parsed as if they are authored by the server.

Polleps:
## Related Issues

closes #1099
closes #658 
closes #913
closes #1024
closes #1107
closes #1101

## Checklist
- [x] Update docs if needed


browser/data-browser/src/views/TablePage/EditorCells/JSONCell.tsx
```diff
@@ -0,0 +1,98 @@
...
+          KeyboardInteraction.EditNextRow,
+        ),
+      ]),
+    }),
+    [isOpen],
+  );
+
+  useCellOptions(options);
+
+  const openDialog = () => {
```
Polleps: 🤔

lib/Cargo.toml
```diff
@@ -17,7 +17,9 @@ name = "benchmarks"
 
 [dependencies]
 base64 = "0.21"
-bincode = { version = "1", optional = true }
+rmp-serde = { version = "1.3.0", optional = true }
+# Needed for migration to bincode v2
```
joepio: msgpack

Issue #998, opened 2024-10-23
Local-first resources - don't require a server when creating a resource


joepio:
# Problem

Creating a new resource requires a lot: an active internet connection, a working server, a resolvable Agent. That makes creating data kinda bothersome, which hinders adoption. It would be nice if Resources can be created and update **local-first**. 

Additionally, it would be nice if resources could be resolved _without a server_. A P2P solution, perhaps something that uses content-addressing (e.g. something like ipfs / iroh) would be swell.

# Solutions

## Solution 1: use Nextgraph

See #1012 

## Solution 2: commit hashes and own schema

- User can create a Commit that does not refer to an HTTP subject. It has no `target`. It's a `RootCommit` (not sure about the name).
- The _signature_ of that commit is the identifier. (or perhaps the `iroh` hash?)
- This `RootCommit` has a `date`, a `publicKey` (linking 
- You can refer to the _initial version_ of this resource using a `did:atomic:RootHash`. So we'll introduce a new `did:atomic` schema for this.
- You can refer to a _specific following version_ of this resource using using a `did:atomic:CommitHash` 
- We still want pretty URLs, in some usecases, of course! So how do we deal with that? How do we map these? 
- We can still resolve using HTTP: `https://atomicdata.dev/cid/aeg8ae9gyg98ea9eahge98gha`

### API (`@tomic/lib`)

Not a lot would change for the user.

- When creating a resource, the `subject` becomes optional, as well as a `server`. The resource that is returned will get a subject through its constructor, by taking the hash of that resource.
- The

### URL schema (resolve method)

If we have local-first resources, these should be able to resolve _without_ HTTP. How would we go about this? If we use NextGraph, we just use their `did` schema. Easy. 

What information should be in the URL? 

- the Schema, e.g. `http` or `atomic:` or `did:atomic`. This describes how the resource is to be resolved.
- The hash / id of the first commit, I think. That's the cryptographic "root"!
- _optional_: the HTTP server (domain), where the resource lives by default. This is mostly there for performance reasons. it allows the client to _quickly_ get the resource from a server
- _optional_: the `path` + `query`. This is the `/my-site/hello` part, that links to the actual content. Not sure if this should be a thing. 

But if we don't, we have a couple of options:

- `did:atomic`, a new schema that we should register. The URLs themselves

## Questions

[...]

PR #1124, opened 2025-11-18
New Document Editor + Yjs type #998

### LINKED ISSUE #720: Consider using yjs / y-crdt for document-related commits
Atomic Commits describe how resources update. I'm happy with part of it, but I'm also having some doubts:

- Merging isn't clear yet and requires an implementation (at very least in the server)
- Atomic Commits doesn't have a formal proof of eventual consistency
- No support for inserts / deletions in text, only replace an entire k/v pair.
- No support for formatting changes
- No cursor positioning
- No great front-end

Doing all these things takes a _lot_ of time. It might be nice if we could just use something like Yjs or Gun to deal with all the state updates.

However, there are some things I still want to keep, ideally:

- It would be nice if the changes themselves are _understandable_. Atomic Commits are pretty simple Yjs seems pretty complicated, which means the user is forced to use a library, and the libraries take a lot of time to build. 
- **signatures** are cool, because we get audit logs
- I want to **persist** the changes and allow re-applying them to build a version history

## What would it mean to use yjs in atomic-server

- Add a `yjs` action to `Commit`s, like we have `set` and `delete`. 
- We keep a yjs document in the front-end for every resource. The `set` and `delete` and `insert` functions utilise these yjs methods. Calling `.save` pushes the state vector to the server in a commit. 
- [yrs](https://docs.rs/yrs/latest/yrs/#quick-start) parses the `yjs` `StateVector` and updates the state in AtomicServer.
- I'm a little concerned as to how big these changes will be. Maybe we should not persist _every_ state/transaction.
- If we switch to NextGraph, we might skip this because it already has a text editor.
- We could replace our editor with blocknote, which also uses yjs #741

### LINKED ISSUE #741: Fancy markdown / document editor -
## Must-have Features

- [ ] Add `Commit` operations for text insertions (real-time collaborative sync)
  - [ ] Implement in JS
  - [ ] Implement in Rust (server)
  - [ ] Make it compatible with codemirror
- [ ] Render Atomic Resources as inline components. Use `[[` syntax like Obsidian.
- [ ] Allow markdown-like syntax to do formatting (e.g. `## Heading 2`) and preview the style

## Cool ideas

- [ ] [Linkifier](https://zulip.com/help/add-a-custom-linkifier)
- [ ] Mentions
- [ ] Custom components
- [ ] Slash commands? Not present in Obsidian

### Codemirror + Markdown

- Use [codemirror](https://codemirror.net/), because:
  - It's reliable, popular, battle-tested
  - Is plaintext
  - Allows for collaboration
- Use markdown because
  - It's simple to read and understand
  - It is highly interoperable

### TipTap

- Has nice notion-style editor
- No native markdown support, but import / export possible using [plugin](https://github.com/aguingand/tiptap-markdown?tab=readme-ov-file)
- Built on Prosemirror
- Can sync, but unclear how / if this can be made to comply with Commits

### [Milkdown](https://milkdown.dev/)

- Markdown first

### BlockNote

- Powered by prosemirror and TipTap, so more batteries-included than TipTap / Prosemirror. 
- Great notion-like UX with slash menu, d&d, live collab
- Can render to markdown (or HTML of course) but markdown is considered lossy.
- Works with Yjs, allows CRDT #720 
- Built in React, which makes things easy for us! 
- Even has cool [comments](https://www.blocknotejs.org/docs/collaboration/comments)

### LINKED ISSUE #1111: Document upgrade path
The current document editor is not so great, it lacks features and is a bit buggy at times.

We already have a nice markdown editor that can do more than the current document editor. We might be better of making a new Document class that only has a markdown field and a title and use the markdown editor instead.

One important thing to keep in mind is that we probably want to move over to some kind of y.js compatible system in the future, Tiptap (our markdown editor) already has supports that (although I'm not sure if it is available in the FOSS version).
Switching to md only will also make adding a y.js representation trivial in the future as we can just add it another property

In order to not break existing documents we should make a new class and migrate old documents to the new version when they are opened, this is probably easier done on the client.
Should we make this automatic or present the user with a button to update and disable editing until the document has been updated or should we do it automatically?

Polleps:
## Related Issues
relates to #998
closes #720, closes #1111, closes #741

## Checklist
- [x] Update docs if needed


Polleps:
@cursor review

Issue #73, opened 2021-01-11
Atomic Plugins


joepio:
Having an easy to use, powerful and flexible abstraction layer is what makes an operating system or runtime successful. Being able to extend an application _at runtime_ is something that only few systems can do. Operating Systems (you can install apps on your laptop or phone) and Browsers are some of the few.

Atomic Server provides a really cool extra bonus: It runs on the web! You can access your atomic server and all its data and apps from any device.

In order to make this plugin abstraction well designed, we should extensively use it internally, during development of (core) features. Some contenders of being Plugins are Versioning (#42) and Collections.

## Some usecases

Let's describe some apps and try to identify which kind of abstractions that would be required to realize this.

### Calendar app

- Host a front-end app (for handling browser requests)
- Read iCal from external calendar, perhaps periodically

### Full text search #40 

- Store and access an index optimized for full-text search (k/v access? Disk access?)
- Re-build / modify index on state changes in main store (access to Commits, as they happen?)
- Host an endpoint for queries
- Host a front-end app 

### Versioning / history #42 

- Host endpoint for queries (fetch version x of resource y, show all commits for resource y, etc.)
- Extend existing resources (not sure if this is necessary, but it might come in handy for adding a `versions` property to existing resources, which increases discoverability)

joepio:
The Mosaic rust project has an interesting WASM powered [plugin system](https://github.com/mosaic-org/mosaic/blob/main/src/wasm_vm.rs).

joepio:
I'd like to make start on running some external code by creating a mock plugin, compiling it to wasm, registering it to an Atomic-Server and running the plugin. I'd like this to be a really minimal plugin, without any store interactions. Perhaps have an endpoint that accepts one argument, which is passed to the WASM function.

joepio:
The minimal example worked fine, but now I'd really like to do things with the Store (e.g. get some data, calculate some numbers, return a new resource with these numbers), and that poses a few challenges:

- Crossing the Wasmer-Rust memory boundary seems necessary for providing a nice API for developers. Ideally, I'd pass the entire `Store`, but doing so is not easy
- If I'm going to use function from Atomic Lib (e.g. `Store` methods), I'm going to have to add WASM as a compilation target. #76.

Maybe I should consider a scripting language such as [Rhai](https://rhai.rs/book/about/features.html)?

joepio:
[...]

PR #1130, opened 2026-01-07
Async + WASM Plugins #73

### LINKED ISSUE #12: Async support for atomic-lib
Since Atomic Data uses a lot of data fetching, some calls might take a while. Sometimes, multiple resources will need to be fetched at once. Instead of executing each request serially, having an parallelized async process for this would minimize execution time. Currently, every call is blocking in `atomic_lib`, and the HTTP library `ureq` is sync, too. Perhaps this will cause issues, but we can change `ureq` for some other HTTP library. 

I'm kind of new to rust, and even more so to async programming, so I need to write down my thoughts on this. First, let's identify which processes might have the highest 'blocking' time. E.g. serializing internal AD3 to JSON. This requires all properties to be fetched, and these might not have been loaded yet.

Now, if we'd start implementing async, a lot needs to happen:

- [ ] ureq has to be replaced by some async HTTP(S) library, preferably with rustls support. Maybe reqwest or [surf](https://crates.io/crates/surf)?
- [ ] a lot of the current functions (serializers that depend on store, get_resource) will have to become async, which will also 'infect' all functions that might use these.
- [ ] the Storelike trait (and perhaps more) need to use [async_trait](https://docs.rs/async-trait/0.1.40/async_trait/)

### LINKED ISSUE #1013: Store key using `SubtleCrypto` API in browser
- We should use the [`importKey`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey) function with `extractable` set to `false` to provide more security.
- [Support](https://caniuse.com/mdn-api_subtlecrypto_sign_ed25519) for `ed25519` is not great, mostly behind flags.

Polleps:
## Related Issues

closes #201 
closes #12 
closes #1013

## Checklist
- [ ] Update docs if needed


joepio:
I think we should consider merging this in even if the plugin API is still highly unstable. I now feel scared to make any substantive changes in the atomic codebase because of how many code is changed to make it async. What do you think @Polleps

PR #1199, opened 2026-05-21
Vector search #1007

### LINKED ISSUE #1007: Indexing with vectors / vector search
LLMs have enabled a new type of search: vector search. Instead of finding a word or string, vector search turns entries and queries into vectors (arrays of numbers). 

RAG search would be [very useful for Atomic Assistant](https://github.com/atomicdata-dev/atomic-server/issues/1049#issuecomment-2754184379).

# UseCases

- Finding things by their semantic meaning (improved search)
- Searching through images, videos, audio, etc. using embeddings
- Finding relevant resources
- Injecting context in an LLM #951 

# Approaches

## Using sled / KV store

I don't think it's possible to do meaningful nearest-neighbor search using sled's KV / BTreeMap data structure. Searching in a KV store is done with `range` queries over lexicographically sorted keys, and I can't see how we can turn a high-dimensional vector in a meaningful key, where close neighbors are also lexicographically close.

## Using an external vector DB

The whole point of AtomicServer is that you _don't need anything else_. I don't want external dependencies, I want it to be a small single binary that you can just run and it gives you all you need.

## [OasysDB](https://github.com/oasysai/oasysdb) (2026: no longer mainteined)

An _embeddable_ vector database in rust. That's the spirit!

## [LanceDB](https://github.com/lancedb/lancedb)

- Embeddable vector search DB, seems to be really fast, also has full-text search (using tantivy). It's not OLTP
- Polle tried it, it was very slow (why?)
- [Not WASM yet,](https://github.com/lancedb/lancedb/issues/230) maybe at some point

## [PolarisDB](https://github.com/hugoev/PolarisDB)

- Zero stars
- Polle tried it, seems really fast
- Unclear if WASM

### LINKED ISSUE #1133: Document editing tools for AI assistent
1. Open a doc
2. Open assistant, ask it to make some eidts
3. It can't read the yjs text

I think it may need some adjusted helper functions. 

Image

### LINKED ISSUE #1184: AI Agent Skills
The AI agent can already edit and read resources so it should in theory be able to do anything a normal user could do. However some actions aren't as straight forward. For example creating a table involves more than just creating a table resource.

To help the AI agent with this problem we should build skills it can read that explain how to perform certain tasks like creating or editing tables and ontologies.

There already exists a standard for skill use called [agent skills](https://agentskills.io/homehttps://agentskills.io/home).
There is however one big issue with this standard: it allows the agent to execute scripts.
This is obviously not possible in a browser environment.

I think we should model our skill feature to be as close to the agent skills standard as possible so it's easier to port existing skills. We could maybe even build an importer that rejects any skill that has a scripts folder.

For now we should store these skills in localstorage but move them to resources in the users private drive when we have that feature.

### LINKED ISSUE #1188: Improved web search for Assistant
Using web search in Assistant is pretty important. These services are typically paid. 

- Brave search, 5$/1000q, 0.5ct/query
- Exa, 7$/1000q, 0.7ct/query and 0.1ct per content request
- Firecrawl, 14$ / 3000 requests

We could let users add API key in GUI to allow this.

### LINKED ISSUE #1192: Assistent setup improvements + e2e spec
When setting up Atomic Assistent, we need the user to jump through some hoops:

- If they want to use ollama, they need to go to the settings page, and enable it there. I think it should happen in-context of the new chat screen.

Goal:

- Assistant works without any setup, in most cases

Some ideas:

- If available, consider using the [Prompt API as a default](https://developer.chrome.com/docs/ai/prompt-api?hl=nl) model (is it good enough though?)

### LINKED ISSUE #1195: Assistent should summarize chat content
When getting cose to token limit, summarize chat and start new context.

Polleps:
## Related Issues

closes #1007 closes #1133 closes #1184 closes #1188 closes #1195 closes #1192

## Checklist
- [ ] Update docs if needed


joepio:
This one is effectively in the did branch #did

PR #1230, opened 2026-07-06
fix(security): guard outbound fetches against SSRF (bookmark/import)

joepio:
## Summary
- `GET /bookmark?url=` and `POST /import` fetched a caller-supplied URL with only a `starts_with("http")` check, no auth gate, and no restriction on the target — letting an unauthenticated caller make the server request internal-only hosts (loopback, RFC1918/CGNAT, and the link-local cloud-metadata endpoint), and via bookmark, read the response back.
- Adds a resolver-level guard (`preflight` scheme/literal-IP rejection + `guarded_resolve` filtering every DNS resolution, including redirects) so DNS-rebinding and redirect-to-internal are covered too.
- Secure by default; `ATOMIC_ALLOW_PRIVATE_FETCH=1` opts back in for intentional internal targets (e.g. LAN federation).
- Reported by Ray Sabee / Whitehat Security (whitehatsecurity.nl).

This is the same fix already shipped in the published [v0.40.2 release](https://github.com/ontola/atomic-server/releases/tag/v0.40.2) (cut from `release/ssrf-v0.40.2`) — this PR brings it into `develop` so future work doesn't regress it.

## Test plan
- [x] New unit tests: IP classification, preflight scheme/literal-IP rejection, escape-hatch behavior, resolver filtering
- [x] End-to-end test: a loopback-resolving domain (`http://localhost`) is refused
- [ ] CI green on this PR

### DIFF
--- a/lib/src/client/helpers.rs
+++ b/lib/src/client/helpers.rs
@@ -8,6 +8,126 @@ use crate::{
     Resource, Storelike,
 };
 
+// ---------------------------------------------------------------------------
+// SSRF guard for outbound fetches of caller-supplied URLs.
+//
+// `fetch_body` fetches URLs that can originate from an unauthenticated caller
+// (the `/bookmark` and `/import` endpoints pass a `url` parameter straight
+// through, and endpoint dispatch returns before the `check_read` auth gate).
+// It previously only checked `url.starts_with("http")`, so an attacker could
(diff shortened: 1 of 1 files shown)


joepio:
Superseded — this commit was cherry-picked directly onto develop in 2d4535fc, so develop already has the SSRF fix. Closing as redundant.

PR #1255, opened 2026-08-05
fix(store): don't fail resources as offline-unavailable before the ClientDb attaches

joepio:
## The bug

`offline-persistence.spec.ts:13` (and part of `sync.spec.ts:266`'s pain) failed on CI only, surviving all retries for days: after an offline reload the drive sat in OPFS with all its properties while the app showed "Offline: resource not available locally".

Call-by-call tracing on CI named it:

```
1012 fetchResourceWithLocalFallback(drive)
1012 waitForServerConnected(5000)   ← no ClientDb attached yet → silence read as a miss
1362 setClientDb(ready=false)       ← attaches 350ms later
2677 setClientDb(ready=true)        ← ...with the drive in it
6013 failResource(drive "Offline: resource not available locally")
```

`initClientDb` derives the agent's database name and unwraps its key before it can attach — landing after React's first render on the fast server-embedded production bundle. A fetch mounting in that window had no database to ask and treated that as "not stored locally". Offline, that failed the resource permanently; nothing ever re-asked. The local vite dev server boots slower than the key unwrap, which is why this never reproduced locally at any CPU throttle.

## The fix

- The OPFS lookup distinguishes **"no database to ask"** (`undefined`) from **"asked, not there"** (`false`).
- The app announces the attach synchronously at boot (`store.expectClientDb()`), so the store knows a database is coming.
- Every path about to fail a resource for lack of local data first waits out the announced attach (bounded, 5s) and asks again. Stores that never announce — Node, SSR, the Sync-page opt-out, Tauri — never wait, and the happy paths are untouched.

Also in here:
- ~76 unguarded `window.store` reads across 21 specs optional-chained (slow boots turned real failures into `TypeError`s).
- `sign-in-without-data` no longer `blur()`s a field that unmounts when sign-in succeeds — it raced its own success path.
- The offline spec keeps its (read-only) diagnostics: the old probe mutated the store before reporting, which masked the real error state.

## Verified

- CI: `offline-persistence:13` passed for the first time (was failed ×3 every run); `tables:323` also cleared. Failure-set diff against the two prior runs shows no new failures attributable to this change.
- Local: lib unit tests 39/39; offline + sync + sign-in sweep ×2 green (except the pre-existing `sync-devices:118` local-env failure).

Still red on CI, **not** addressed here (independent, pre-existing): the totals/aggregates family (`row-actions:118`, `table-templates:208`, `dashboard:419`) and `sync:266` — being investigated on a separate branch.

---

> [!NOTE]
> **Medium Risk**
> The ClientDb attach grace period changes core offline fetch behavior and could delay or alter failure timing; the rest is mostly branding, docs, and test stability with broad but low-risk touch points.
> 
> **Overview**
> **Offline / local cache:** Fixes drives and other resources showing as permanently unavailable offline when OPFS already had them. `initClientDb` now calls `Store.expectClientDb()` before the async attach, and fetches that would fail a resource wait up to 5s for the ClientDb instead of treating “no DB yet” as a local miss. Local hydration is split into `hydrateFromLocalDb` (returns `true` / `false` / `undefined`) and retried on offline failure paths. The offline-persistence e2e is tightened (read-only diagnostics, optional `window.store` during slow boot).
> 
> **Brand & legal:** Adds a `brand/` package (source SVGs, separate `brand/LICENSE`, `TRADEMARKS.md`, root LICENSE trademark notice, README section). `brand/generate.mjs` regenerates favicons and launcher icons across apps; marks move from ring-and-orb to the open “sweep” design. Boot splash (`index.html`), `Logo.tsx`, and `logo.svg` use the new lockup with an animated spinning “o”.
> 
> **Product copy:** About page tagline, intro, and feature list are rewritten; i18n `.po` files updated accordingly.
> 
> **Tests:** Widespread e2e use of `window.store?.` to avoid races before boot; sign-in e2e drops post-fill `blur()` that raced dialog teardown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 25be60a76f4b37334c75ca702dd65787b06eb174. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joepio:
Landed via #1257, which contained this branch's commits (develop now includes them verbatim — the diff here is empty). Root cause and validation are in #1257's description.

PR #1256, opened 2026-08-05
fix(tables): re-read the totals when a member changes outside the save path

joepio:
## The bug

The CI-only "em-dash totals" family: `row-actions:118`, `table-templates:208`, `dashboard:419` (and `tables:323`'s row-visibility variant) failed with the totals label rendered and the value stuck at "—", never recovering. Not reproducible locally at any CPU throttle.

Instrumented CI runs (the `[agg]` rig, added and later removed in this PR's history) peeled back **four layered causes**, each visible only after fixing the one above it:

1. **No re-read on out-of-band change.** `useTableAggregates` refreshed only on `ResourceSaved`/`Removed`. A member changed by drive re-sync, another tab, or a live push emits only `ResourceUpdated` — the very signal that keeps the grid's own membership live — so the totals froze on the last read, forever.
2. **Reload durability race.** A row's post-ack re-persist commits with `Durability::None`; a reload inside the 1s flush window rolls it back — member indexed (earlier flush), newest value gone.
3. **Debounce starvation.** The post-reload drain/re-sync emits a save every ~50–150ms for many seconds, and every event reset the 500ms debounce: the timestamped trace shows 15s of triggers with **zero** reads between them.
4. **Worker-queue latency.** With the starvation fixed, reads start through the storm — and none complete for 10–17s, then five finish within 60ms, all correct: FIFO-queued behind the re-drain's writes, each paying a full index rebuild (the tracked OPFS write-amplification issue).

## The fixes

- `useTableAggregates` also subscribes to `ResourceUpdated`, scoped to the query's membership pair and gated on `lastCommit` advancing (its own reads hydrate members and re-emit the event — an unguarded refresh polls forever). Dashboard blocks share the hook.
- The refresh debounce gets a **max-wait** (1.5s): a pending re-read can't be pushed back indefinitely by a continuous event stream.
- Failed aggregate reads now `console.warn` instead of vanishing — a totals row frozen by a swallowed error is part of how this stayed undiagnosed.
- `reloadGrid` (e2e) flushes the ClientDb before reloading — `pendingDirtyCount === 0` proves server ack, not OPFS durability — and `tables` "fast row entry" stops using a bare `page.reload()` on forty fresh rows.
- The totals assertions get budgets matched to the measured convergence (~18.5s on a loaded runner → 30s), each commented with the cause and the condition for shrinking them again.

## Verified

- CI: the whole family green (run 31051697392) — first run with zero totals failures; only `sync:266` (independent, pre-existing) remains red on the branch.
- Local: 30/30 across row-actions, table-templates, dashboard, aggregates, tables and the offline family on the cleaned branch; lib unit tests 229/229.

## Follow-up worth its own issue

~15 **already-acked** rows re-drain and re-persist on every reload (`pendingDirtyCount` was 0 before it). That storm is why the worker queue is seconds deep at all; fixing it (or the write amplification underneath) lets the 30s budgets shrink back.

Contains #1255's commits; merge after it.

---

> [!NOTE]
> **Medium Risk**
> Changes to aggregate refresh and ClientDb attach timing affect table footers and dashboards under load; the rest is mostly branding, docs, and test stability with limited runtime risk.
> 
> **Overview**
> **Table totals** no longer stick on "—" after reload or sync storms. `useTableAggregates` re-reads when table members change via `ResourceUpdated` (not only saves), with `lastCommit` deduping and a **1.5s max-wait** on the debounce so continuous post-reload events cannot starve reads. Failed reads log warnings instead of failing silently.
> 
> **Local DB boot**: `initClientDb` calls `store.expectClientDb()` before attach so early fetches wait instead of permanently failing offline with "resource not available locally".
> 
> **Brand**: New `brand/` source marks, `generate.mjs` fan-out to favicons/icons, `TRADEMARKS.md` + separate `brand/LICENSE`, README licence section, and **open-sweep** mark on splash (`index.html`), `Logo.tsx`, and lockups.
> 
> **E2E**: `reloadGrid` flushes ClientDb before reload; optional chaining on `window.store`; longer timeouts for aggregate assertions; offline-persistence diagnostics without mutating cache; sign-in test no longer races dialog teardown on blur.
> 
> **About** page copy and feature list aligned with README/marketing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83b33878a6de91dafded9360f8b2ebeeed7af6e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joepio:
Landed via #1257, which contained this branch's commits (develop now includes them verbatim — the diff here is empty). Root cause and validation are in #1257's description.

PR #1288, opened 2026-08-19
Bug fixes in 1286

joepio:
This pull request contains changes generated by a Cursor Cloud Agent

<sup>Cursor generated this PR to address Bugbot findings on PR #1286. If that PR is closed, this PR will be closed automatically. Configure in [dashboard](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/browser/data-browser/src/components/ResourceContextMenu/index.tsx
+++ b/browser/data-browser/src/components/ResourceContextMenu/index.tsx
@@ -52,7 +52,11 @@ function runAction(action: ActionDefinition, ctx: ActionContext): void {
   }
 }
 
-function reportActionError(action: ActionDefinition, ctx: ActionContext, e: unknown): void {
+function reportActionError(
+  action: ActionDefinition,
+  ctx: ActionContext,
+  e: unknown,
+): void {
   const detail =
     e instanceof Error && e.message
       ? e.message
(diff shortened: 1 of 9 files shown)


joepio:
Closing: this branch no longer contributes anything. `git diff fix/sync-nodes-index-and-ui cursor/identified-bug-fixes-ec06` is empty.

All three fixes landed on the base branch independently, in a more complete form:

- **Action error label** — `381eaf6d` also wraps `action.label(ctx)` in a try/catch with an `'Action'` fallback, since resolving the label reads the resource and can itself throw. A label that fails must not swallow the error the toast exists to report.
- **Count repair** — `381eaf6d` added the same surplus guard, then `bd83cea6` reverted the whole count probe: it fired a query per collection and broke the invariant `query-drive-filter.spec.ts:100` exists to assert. The real cause was server-side, fixed in `6c83924f`.
- **Tauri drive restore** — `381eaf6d` restores in the `did:` branch instead (`wasExplicitlyChosen() ? get() : getLocalServerOrigin()`), rather than persisting the foreign origin. Dropping the `!isRunningInTauri()` guard as this branch did would store that origin on Tauri and clear the explicit marker, reinstating the bug where one visit to an `https://…/drive/…` entry left the app booting against that server.

The merge I pushed here (`badfc5cb`) resolved every conflict to the base version and produced a tree identical to it, which is what made the no-op explicit. Closing collapses the stack to #1286 against `develop`.

Issue #277, opened 2022-01-07
Consider supporting OIDC / SSO / Oauth


joepio:
Atomic Data introduces a new self-sovereign public key based authentication mechanism. This has its merits (users are in full control of their identity, instead of servers), but pretty much everyone out there already settled on a standard: Open ID Connect. Supporting this helps achieve a few goals:

- Sign in with your _existing_ OIDC account (google, facebook, or your companies SSO server). No need to create a new account!

Libraries:

- https://github.com/ramosbugs/openidconnect-rs
- https://github.com/jofas/actix_oidc_token

- [ ] Setting up an account would require email #276 
- [ ] Posting new data would require something different than the Commit endpoint.
- [ ] Parse / set tokens

FreeSynergy:
How far is this feature? Will it get implememted?

joepio:
Hey @Lord-KalEl, i won't be using my existing (huge) PR that implemented this, but we're starting from scratch. We're planning on implementing this in ~the next year~ 2026.

FreeSynergy:
> next year.

You mean this year … ;)

2026 … (I hope … lol)

joepio:
lol yeah sorry

FreeSynergy:
I hope you will implement SCIM … this is the newes standard … 

https://scim.cloud/

And thanks … for your work …

alexanderadam:
> i won't be using my existing (huge) PR that implemented this, but we're starting from scratch.

Just to be sure that I don't miss this: there's no open PR or branch that one can track yet, right?

joepio:
> Just to be sure that I don't miss this: there's no open PR or branch that one can track yet, right?

Nothing worth pointing to, but I think I should re-prioritize this @alexanderadam

PR #1310, opened 2026-08-27
Retarget OIDC / OAuth after DID and local-first

joepio:
## Related Issues

Related: #277 (retarget, do not close)

## Checklist
- [x] Update docs if needed

## What this is

A reconsideration of #277. Planning only, no protocol or server code.

## Decision

OIDC-only login is a CA: root Agent on the node, browser-generated session Agent, compact SessionCert. Stock authorization-code OIDC; token-exchange into the cert, not Bearer on `/commit`. Export + unlink so Google is reversible.

## How verifiers check session keys

Signature and rights split. `commit.signer` is the session DID; `effective_agent` remaps to the root after `SessionCert::verify`. `check_rights` does not grow a cert case.

1. Ed25519 of the commit / AUTH message under the **session** pubkey (today's `validate_signature` / `check_auth_signature`).
2. Compact SessionCert: session pub matches signer, root Ed25519 over the blob, `t` in `[notBefore, notAfter]` (`createdAt` on commits so history stays valid; AUTH timestamp on live requests).
3. `check_write` / `admit_drive_write` / genesis `write` insert / AUTH's returned `ForAgent` use the **root** DID.

HTTP GET, WS, and Iroh inherit the remap from `get_agent_from_auth_values_and_check`. No IdP at verify time. No cert means today's path.

## Known weakness, stated in the doc

`createdAt` is signer-chosen and no code path rejects an old timestamp, so a stolen session key is only bounded on live-ingest paths (HTTP, WS, Iroh COMMIT get a `now <= notAfter + skew` check). Replica and catch-up ingest stay exposed until auth-checkpoint receipts land. See "Revocation and the `createdAt` problem".

## What an OIDC user gets

One click, nothing to save, new device with nothing to transfer. Sharing is by invite link in v1 (ACLs hold opaque root DIDs); email lookup is a follow-up. Deprovision lags to `notAfter` (24h default). Identity is per node. Flutter later. See "After login: what an OIDC user can and cannot do".

## Impact

Absent IdP config, nothing changes. The costly work is SessionCert in `validate_signature` + AUTH, plus the node minting the personal Drive on first visit (derivation needs the root private key). The OIDC plugin is ordinary. Operator sets issuer/client/secret; end users only click the button.

User-facing docs describe this as planned, not shipped.


joepio:
Closing as a duplicate of #1326 during the requested PR cleanup. The current heads have identical Git file trees (c74aff5cdb572dec79f953a80b4004b87d93a114), so keeping both PRs would duplicate the same changes. Continue review in #1326; reconcile the differing PR descriptions with the intended control-plane OIDC design there. The branch is retained.

PR #1326, opened 2026-09-01
Session certificates: a root Agent certifying a short-lived session key

joepio:
## Related Issues

Related: #277 (retarget, do not close). Supersedes the planning-only #1310.

## Checklist
- [x] Update docs if needed

## What this is

The protocol half of `planning/oidc-oauth.md`, plus the decision document itself.

A **root** Agent is the person: its DID is what `write` lists name and what `createdBy` records. A **session** Agent is one browser — a fresh Ed25519 keypair whose private half never leaves that device, and which the root signs a short-lived certificate over. The session key signs commits and authentication proofs; every rights check is answered for the root.

OIDC login itself is **not** implemented. There is no relying party, no `/oidc/*` route, no provider button, no issuer config. Absent a certificate, every path is byte-identical to before — asserted in the tests, not assumed.

## What ships

**`SessionCert`** (`lib/src/session_cert.rs`, 145 bytes):

```text
offset  size  field
0       1     version        0x53
1       32    sessionPubKey  Ed25519 raw
33      8     notBefore      i64 unix ms LE
41      8     notAfter       i64 unix ms LE
49      32    rootPubKey     Ed25519 raw
81      64    signature      Ed25519 by the root over bytes [0, 81)
```

Mirrored byte for byte in `browser/lib/src/session-cert.ts`, proven by a golden fixture the Rust side generates and both sides check, so the two can only ever agree or both fail.

**One helper, every call site.** `effective_agent(signer, cert, t)` is the only place the certificate case lives. Without a certificate it returns the signer; with one it verifies and returns the root. `check_rights` grows no cert case — it keeps comparing a DID to `read` / `write` / self.

**Commits** carry the certificate as an optional `sessionCert` propval *inside* the signed JSON-AD, so it is bound to the mutation and cannot be stripped without breaking the signature. `validate_signature` verifies it against `createdAt`, so history stays verifiable after the window closes. `commit.signer` still records which device signed; rights, `createdBy`, the genesis `write`-list insert and the drive-enrollment gate all use the root.

**AUTH** carries it as `x-atomic-session-cert`, or as `auth/sessionCert` on the Authentication resource. HTTP, the `atomic_session` cookie, WebSocket `AUTHENTICATE` and the Iroh auth-back all reach `get_agent_from_auth_values_and_check`, so one change covers every transport. `sync::engine` mints no Agent resource for a session DID: it is a public key with a lifetime of hours, and publishing it forever is the opposite of the point.

**Vocabulary.** `sessionCert` and `auth/sessionCert` are declared Properties in `lib/defaults/default_store.json`, and the `Commit` class recommends the former.

**Client plumbing.** An `Agent` may carry a certificate; when it does, `CommitBuilder` signs it in and `createAuthentication` / `signRequest` attach it.

## The version byte is 0x53, not 0x01

The draft said both "version 0x01" and "domain-separated by version", and those cannot both hold. `GenesisCert` v1 also starts `0x01`, and a genesis certificate is a signed 81-byte payload whenever `parent` and `drive` come to 19 bytes together — exactly a session certificate's signed length. The same signature would verify under either reading, so a genesis signer would appear to have delegated to whatever bytes sat at the session-key offset. Distinct leading bytes make the two payload languages disjoint at byte zero. Pinned by a test on both sides.

## Known weakness, deliberately

The window is checked against `createdAt`, which the signer picks, and no path in this codebase rejects an *old* `createdAt` — on purpose, since a bulk reconcile legitimately re-sends old commits.

So every path that already validates timestamps (HTTP `/commit`, WS `COMMIT`, Iroh `COMMIT`) now also requires `now <= notAfter + skew`. The sender is online on those paths, so an honest client never notices, and a stolen key stops working at `notAfter`.

Replica ingest and Iroh catch-up skip timestamp validation by design and therefore **stay exposed**: a backdated commit can still enter through a peer that accepts it. That closes with auth-checkpoint receipts, not here. There is a test that pins both halves of this — the live path refuses the backdated commit, the replay path still accepts it, because history must stay verifiable.

## Tests

| Suite | Result |
| --- | --- |
| `atomic_lib` (12 cert, 4 AUTH, 7 commit end-to-end, plus existing) | 716 passed |
| `atomic-server` (3 new, through the real handlers) | 229 passed |
| `@tomic/lib` | 849 passed |

The commit tests go end to end against a store: a certified session key writes with the root's rights; the same key without its certificate is a stranger; a certificate from a root that lacks `write` is still refused (identity is not authority); genesis puts the root in the `write` list and never the session DID.

The three server tests drive the actix handlers, covering the two parses the library tests cannot reach: a posted commit is parsed back by `Commit::from_resource`, which has to read the new propval, and a request carries the certificate in a header `get_auth_headers` has to pick up.

**One of them failed when written, which is why they were worth writing.** A commit carrying the certificate made the server try to *fetch* `https://atomicdata.dev/properties/sessionCert` to learn the propval's datatype, because the URL had no Property behind it — a 500 on any node, and an unauthenticated outbound request on every such commit. Fixed by declaring the property; only propvals carried on the commit resource itself need this, which is why `genesis`, set inside the Loro doc, never did.

**No end-to-end (Playwright) coverage, and none is possible yet.** Those specs drive the real app in a browser, and nothing in the data-browser ever sets a certificate: there is no sign-in button and no `/oidc/*` route. That arrives with steps 3 and 4.

Two things could not run in the authoring sandbox and need CI: `sync::tests::peer_sync_tests::pkarr_discovery_and_iroh_sync` needs an external pkarr relay the egress policy blocks, and the data-browser build needs `codeload.github.com`, blocked the same way. The `-D warnings` workspace clippy the pre-commit hook runs was executed manually and passes.

## Not in this PR

Steps 3 to 5 of the plan: the relying party on the node (issuer config, `/server` advertisement, `/oidc/start`, `/oidc/callback`, `POST /oidc/session`, wrapped roots), the browser sign-in flow, and export + unlink. `planning/oidc-oauth.md` marks which step is built.

Noticed while testing, not touched here: three tests in `lib/src/db/test.rs` open a database at a hardcoded `.temp/db/<name>` path instead of randomizing like the rest, so leftover state from an interrupted run makes them fail until `.temp` is cleared.

https://claude.ai/code/session_01UxNgUQWzExgQkmKhNg4TxB


joepio:
**Downstream (atomic-saas builds against this commit)** failed on `922c6f7f`, and it is not this PR's.

The portal TypeScript build, which failed on the previous head, passes now. The job got as far as the atomic-saas Rust suite, which reported **445 passed, 1 failed**:

```
vault_handler_tests::recovery_requests_are_audited_without_the_envelope
assertion `left == right` failed: one line per request:
  left: 3
 right: 5
```

**Why it is not this PR's.** The test drives atomic-saas's own `/api/recovery-secret` endpoints and counts `saas::recovery_audit` lines captured through a thread-local `tracing` subscriber. This PR touches commit signing, request authentication and the default store; nothing in that path. The two lines missing from the capture are the two *successful* requests, while the rate-limited, unauthorized and failed ones were all recorded, so the assertion is about which audit writes had landed when the sink was read, not about anything the diff changes.

**Reproduction attempt.** Built atomic-saas at `3814a23` (the commit CI checks out) against this branch and ran the full binary suite six times: 446 passed, 0 failed, every time, and the named test also passes in isolation. The 446 matches CI's 445 + 1, so it is the same test set, not a different feature selection.

**No fix to port.** There is no atomic-server change that would address it, and the test is in another repository, so fixing it here is not possible. The direction I would suggest to whoever owns it: the test is `#[tokio::test]`, so the runtime is single-threaded and the thread-local subscriber assumption is sound for inline work, which points at the success-path audit being written outside the inline path rather than at the subscriber. Asserting on the audit lines that must be present, rather than on an exact count, would also make it robust without weakening what it checks.

**Re-run deferred.** Re-running the failed job was refused with `403 This workflow is already running` — the Mancave CI job in the same run is still going. I will spend the one re-run once that run completes, and keep watching until this PR is green.

Everything else on this head is green, and the branch merges cleanly into `develop`.

---

PR #1330, opened 2026-09-02
Revert "fix(populate): re-seed built-in defaults on Db open via a fingerprint (#1323)"

joepio:
## Why

Every `develop` push since 4aa5afe4f (#1323) fails the full e2e suite on the same tests, 3/3 attempts each: `calendar.spec.ts` "add an item on a day, and it persists", `documents.spec.ts` "shows a collaborator's ephemeral cursor position", `tables.spec.ts` "fast row entry". All three read state back through a reload or a second session. The run right before that commit (3f2ca1875) was green, and the commit is the only change in that window.

## What I could and could not reproduce

- Locally, with server binary + WASM + frontend all built from `origin/develop`, the ephemeral-cursor test failed once (page 2 got the document title but not the Loro text; the server's snapshot and commit chain never contained the typed text) and passed on later runs. With #1323 reverted it passed every time. Swapping only the server or only the WASM did not reproduce either, so it is timing-sensitive and needs the CI runner's load to show reliably.
- The `[full-e2e]` marker on this commit makes CI run the full suite on the branch. This PR is a draft until that run is in.

## Follow-up

The feature (re-seeding defaults on open) is worth keeping; the revert is to get `develop` green while the mechanism is found. The suspicious part is that `Db::init_redb_opfs` now runs the seed on every WASM open whose fingerprint mismatches, inside one `Durability::None` batch, and the JS-side seed in `initClientDb.ts` trusts the WASM defaults on a first visit.

### DIFF
--- a/browser/e2e/tests/test-utils.ts
+++ b/browser/e2e/tests/test-utils.ts
@@ -2011,3 +2011,5 @@ export async function acceptInvite(page: Page) {
     40000,
   );
 }
+
+// CI cache-bust 2026-09-02: forces a real e2e execution for this branch.
--- a/lib/src/db.rs
+++ b/lib/src/db.rs
@@ -210,10 +210,6 @@ pub struct ReplicationTarget {
 
 const REPLICATION_PREFIX: &str = "replication:";
 
-/// `Tree::PluginMeta` key holding the fingerprint of the built-in defaults
-/// last seeded into this store. See `populate::bootstrap`.
-const DEFAULTS_FINGERPRINT_KEY: &[u8] = b"defaults_fingerprint";
-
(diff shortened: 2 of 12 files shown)


joepio:
Closing: an uncached full-suite run of this branch (33627864052) still fails the calendar-persist and fast-row-entry specs, so reverting #1323 does not make develop green. Worse, building this branch wrote a stale `atomic_lib` (without `repopulate_defaults`) into the shared cargo target cache, which then broke compilation for concurrent runs on develop and other branches (E0425 in the jsTestIntegration slim build). Deleting the branch so it cannot poison the cache again; a cache reset follows in #1329.

PR #1335, opened 2026-09-02
fix: never queue Commit resources in the outbox; define server/homeDrive

joepio:
From a sign-in on staging.atomicdata.dev (browser log in the session).

**Outbox.** Three `https://staging.atomicdata.dev/commits/<sig>` entries drained on every reload and were refused with `Commits cannot be edited.` — Sync page permanently on *Changes pending*. The Loro local-edit subscriber skipped `did:ad:commit:` subjects but not the older HTTP `/commits/<sig>` form. Now: `isCommitSubject()` covers both; `markDirty` ignores them; `hydrate` drops ones an older build persisted; the server message is a terminal (drop) error. Tests for each.

**Server.** `GET /server` with `ATOMIC_HOME_DRIVE` set (staging has it) answered 500 for `Accept: application/json` / `ld+json`: `server/homeDrive` was never added to `default_store.json`, so plain-JSON rendering tried to fetch the property from atomicdata.dev (404). JSON-AD was unaffected, so the app itself worked. Property added (also listed under the `Server` class `recommends`); the regression test fails without the JSON change and passes with it.

Not touched: why the drives did not show on the first sign-in with the legacy agent — the second attempt loaded them, and nothing in the sign-in path waits on the outbox.

https://claude.ai/code/session_019asLKBrBWY5ovyeCgtmdSd


joepio:
Added a third fix (same staging session): a returning account on `app.staging` whose IDB agent record predates `Agent.privateDrive` threw *Cannot work out this account's private drive: its key signs non-deterministically…* from `createDrive ← ensurePrivateDrive`. The Store now falls back to the `personalDrive` pointer on the Agent resource (local copy, then server) and caches it for the session. Two lib tests.

PR #1384, opened 2026-09-07
Security audit: fix sync, authentication and rights gaps, bump advisories

joepio:
## Related Issues

No tracked issue; the findings are in `planning/security-audit-2026-09.md` (with a status section per finding). Companion PR: ontola/atomic-saas#55 (same branch name).

## What changed

**Critical**
- Removed the committed editor config carrying live API keys, plus the tracked pnpm-store and vitest cache files. The Context7 and Brave keys still need revoking at the vendors.
- `POST /iroh-sync` requires a signed agent with write on the drive (or the policy's leave for a new drive) and only answers POST; dial-side owner trust is confined to the drive dialed for.
- `SYNC_PUSH` checks every entry against the admitted drive.
- `https://host/agents/{key}` subjects bind the path key to the signing key; agent subjects on other hosts are refused instead of fetched.

**High**
- `check_append` no longer trusts the new resource's own `write` array; parentless creation is limited to cert-bound DIDs without a parent, drive-stamped children whose drive the agent may append to, and top-level Drives by the node's agent. Regression tests included.
- Request origin comes from `Host` / `X-Forwarded-*` only for the configured domain, its tenants, or loopback (`*.localhost` included).
- `/forget-peer` requires write on a drive the peer was paired for; known peers record those drives.
- Iroh AUTH is bound to the responder's node id (`drive#nodeId`); unbound proofs are honoured only from paired peers.
- Desktop and Android bind the embedded server to `127.0.0.1`; the desktop `devtools` feature is opt-in.
- The data-browser pins the portal a device token was issued by, validates a node-reported `portalUrl`, and rejects `javascript:` hrefs.

**Medium / Low**
- Blob responses are hash-checked; meta tags and the plugin UI query string are escaped; upload and search limits are bounded; LanceDB literals are quoted; rendition parameters are quantized; `config.toml` is written 0600; path-only auth signatures are refused; the signer's Agent resource is created only after acceptance; genesis certs must match the document; plugin zips and bookmark bodies go through the SSRF guard with size caps; ephemeral relays require a subscriber; ACME errors instead of panicking and certificates are renewed daily on disk; plugin UI reads use the calling agent; `Agent`/`SharedConfig` redact secrets in `Debug`; the Flutter bridge deletes the DB only on corruption; CI workflows drop the persisted PAT and use per-job permissions.
- Sloppy code: `ValueComp` comma-case, unawaited `resource.save()` calls, the CLI agent race, dead code and duplicates.
- Dependencies: wasmtime 47.0.4 (sandbox escape), h2 0.4.19, quinn-proto, rustls-webpki 0.103.13; `@tiptap/*` 3.30, `@modelcontextprotocol/sdk` 1.30 and transitive build tools (`pnpm audit --prod` is clean). The `@tiptap` bump left two `prosemirror-model` versions in the lockfile, which broke `pnpm typecheck` in data-browser; `pnpm dedupe` collapsed them (and other duplicates to versions already present).

**Withdrawn from this PR**
- Drive-scoping of `did:` resources in watched queries (C17). The first version compared a resource's `drive` stamp with the query's `drive`, but for a `did:` query that field is the queried subject itself rather than the drive root, so legitimate rows were excluded and the query-aggregate tests failed. Reverted; the finding is back under "not fixed" in the report with what a correct fix needs.

**Ported from develop's current head and from #1386 (not audit findings)**
- Lint fixes for code develop gained in #1389 and #1390 (missing blank lines, shadowed names in the new e2e specs), and a 60-second budget for the data-browser test that runs the React compiler over every TSX file, which exceeds vitest's default on a two-core hosted runner. The two lib upload integration tests now create the agent's private drive instead of uploading under the server's root URL, which no longer exists as a resource.
- The `ClientDb` cold-start fix from #1386 (with its unit test): the leader election no longer steals its own OPFS lock and logs a ghost-leader warning when the tab already holds the lock and its worker is still importing WASM. That warning tripped the browser-diagnostics gate from #1382 on every fresh page on a loaded runner, which is what kept the light e2e suite red on `develop` and here. It no-ops once #1386 lands.
- The FOSS logout guard from #1386 (with its unit test): signing out no longer posts `/api/logout` to a self-hosted node's own origin, which answers 405 and failed the sign-out smoke spec on the same gate. `hasManagedApi()` here is written against this branch's portal sources (linked portal, build-time API base, remembered portal, compiled-in portal URL).

## CI note
Before the ports above, the e2e job on this branch failed only on failures that are red on `develop` since #1382 (the `[ClientDb]` lock-steal warning tripping the browser-diagnostics gate, and the ontology dialog spec on Mancave); see the comment thread. With the first port the hosted run went from 11 failing smoke specs to 1 (the logout 405); the second port targets that one.

## Checklist
- [x] Update docs if needed

https://claude.ai/code/session_014tcSpZhjA3SgzcEfRdUr67

### Follow-up review fixes

Integrated desktop restoration PR #1391 to verify the combined protocol behavior. Workspace inspection now signs `drive#nodeId` using the same helper as normal peer sync, so fresh devices pass the new node-bound authentication check.

Rejected cross-drive `SYNC_PUSH` entries no longer persist snapshots before validation. Accepted state is stored through the existing atomic resource/snapshot transaction, and failed persistence is not counted as imported. A regression checks that rejected state leaves no snapshot and cannot contaminate a later valid import.

Validation: both regressions failed before the fixes and pass afterward. The native sync run passed 171 tests with one ignored and one presence-test failure under parallel execution; all 15 Iroh end-to-end tests passed when rerun with the documented `--test-threads=1`. `cargo fmt --all -- --check` and `git diff --check` pass. No local Dagger run was restarted.


joepio:
CI status on `989143c` (rebased onto `develop` at `584d696`):

The Mancave e2e job (`--playwright-mode light`) fails on 11 `@smoke` specs. All of them fail the same way on `develop`'s own run of #1382 (run 34198467931, 37 to 46 failures per shard), so they are not this PR's:

- 10 specs (dashboard, drive-deeplink, invite share menu, chatroom, delete resource, folder, kanban, meetings, pairing dialog, search, second-device, sync, tables) fail only the new browser-diagnostics gate from #1382, on the warning `[ClientDb] no leader-announce in 2000ms; stealing OPFS lock from suspected ghost leader`. The steal succeeds and the tests' own assertions pass; the 2 s election window is what the loaded runner misses. The same warning fails `shortcuts` and `sign-in-without-data` on `develop`.
- `ontology › Create and edit ontology` (the New Class dialog's Save button not found) is in `develop`'s failure list for #1382 as well.

The two failures this PR did introduce were fixed in `3fa6d34` (the request-origin check rejected `atomic.localhost`) and `989143c` (the new session test's api mock).

No fix for the diagnostics failure exists on `develop` yet. Proposed patch, not pushed here to keep the PR to its scope: in `browser/lib/src/client-db.ts` log the successful lock steal at `console.info` instead of `console.warn` (the gate only collects warnings and errors), or raise `LEADER_ELECTION_WAIT_MS` under CI. Happy to push either into this PR if that is the preferred route.

---

joepio:
GitGuardian: the string it flagged in `lib/src/agents.rs` (commit `de72563`) is the Ed25519 test fixture that `lib/src/commit.rs` has used in `store_with_known_agent` on `develop` for a long time, not a live credential. The new copy is removed at `0f152a0` (the redaction test now generates its key at runtime). The historical occurrence in this branch's commit will keep the incident open until it is marked as a test credential in the GitGuardian dashboard, or the branch is squashed on merge.

---

joepio:
CI on `ca8c847` (hosted runner, light mode): 11 `@smoke` specs failed, all on the browser-diagnostics gate and all on the same warning, `[ClientDb] no leader-announce in 2000ms; stealing OPFS lock from suspected ghost leader`. No other diagnostic appears in the log; the specs' own assertions pass.

#1386 carries a fix for exactly this (a tab that already holds the lock while its worker is still loading no longer steals from itself), and its hosted CI run is green. Ported that change plus its unit test into this PR at `2aa33bb`; it no-ops once #1386 merges. The ontology dialog spec is Mancave-only and was not in this run.

---

joepio:
Hosted run on `2aa33bb` (with the ClientDb port): 16 of 17 smoke specs pass, up from 5. The ghost-leader warning is gone from the log entirely.

The one remaining failure, `sign in with secret, edit profile, sign out`, is again the diagnostics gate, this time on a browser error: `POST /api/logout` answered 405 by the FOSS test server's own origin. #1386 fixes that too (it skips the SaaS logout when no control plane is configured); ported it with its unit test at `f6e3e1c`, adapted to this branch's `api.ts`.

While validating that locally I found `pnpm typecheck` in data-browser failing on this branch: the `@tiptap` bump had left `prosemirror-model` 1.25.7 and 1.25.11 side by side. CI does not run typecheck, but it is this PR's regression, so `2c4d2b5` dedupes the lockfile (typecheck, lint, unit tests and the vite build pass; frozen install is consistent).

---

PR #1387, opened 2026-09-08
API plugins using localthought/integration-proxy

michielbdejong:
Adds browser-owned integrations from a `localthought/integration-proxy` server alongside the existing bundled plugins.

Users can choose the server under **Settings → Integration** without rebuilding the frontend. The form follows the Ollama settings layout, validates HTTPS/loopback origins, and supports resetting to the deployment default (`VITE_INTEGRATION_PROXY_URL`, falling back to `https://localthought.io`). Saved connections are isolated by proxy origin; existing browser connections migrate only when their origin and owner match.

Proxy integrations have an accent border and a “Via integration proxy” label. The original GitHub, Notion, Clockify and MT940 cards remain available independently, including when the proxy is unavailable. Existing provider implementations, Notion OAuth protocol, certification metadata and shipped-bundle regression tests remain in the repository.

The proxy flow discovers a catalog, connects accounts with PKCE, imports paginated records through Syncables/WASM, and previews changes before applying them locally. Generic metadata-driven traversal and request-limit improvements remain included.

Validation:
- LocalThought/settings tests, restored provider fixture suites, provider UI tests and certification checks pass.
- Frontend typecheck, lint and formatting pass.
- Browser verification covers settings search, save/persistence, URL validation, reset, proxy catalog selection, and the original GitHub setup alongside the proxy card.
- Restored OAuth and provider sandbox tests run against the actual shipped bundles.

Live provider tests requiring credentials and the full browser E2E suite were not rerun for the scope-restoration change.

## CI scope

General CI/cache/runtime hardening, workflow changes, and certification tooling improvements have been split into #1424, independently based on `feat/plugin-model`. This PR retains only CI additions required to build and test the integration feature. No checks were disabled. CI is being rerun on the narrowed branch.


michielbdejong:
**CI red: `Main (GitHub) / CI` — not this PR's, pre-existing on `feat/plugin-model`**

The failing step is the production frontend build (`jsBuild()` → `vite build` in `data-browser`), which errors with `[UNRESOLVED_IMPORT]` for every `../../../../../integrations/<name>/*` import in `browser/data-browser/src/chunks/PluginRuns/*.tsx` — now including `ConnectPets.tsx`'s imports of `integrations/pets/schema` and `integrations/pets/plugin.js?raw`.

This reproduces identically on this PR's own base commit, `feat/plugin-model`@`3563a7ee24` (before any of my changes), for the pre-existing `mt940`, `clockify` and `github-issues` integrations: [run 34218651424](https://github.com/ontola/atomic-server/actions/runs/34218651424/job/102047631007) fails with the same `Could not resolve '../../../../../integrations/mt940/schema'` / `.../clockify/atomic` / `.../github-issues/plugin.js?raw` errors. Every push to `feat/plugin-model` back to late August has this job red for the same reason, so it isn't a flake and it isn't specific to `pets` — I didn't spend the one re-run on it since a fresh run would reproduce deterministically.

**Root cause**: `.dagger/src/index.ts`'s `jsBuild()` (used by the production `vite build` / e2e / netlify-preview pipeline) mounts only `browser/` into the container (`.withDirectory('/app', browser)`, ~line 1048). It never mounts the repo-root `integrations/` directory, so any relative import reaching outside `browser/` into `integrations/` can't resolve inside that container — even though it resolves fine locally (full checkout) and under `tsc`. The JS-certification stage (`integrationCertification`-ish, ~lines 452/472) already does mount it (`.withDirectory('/integrations', this.source.directory('integrations'))`), so this is just a gap in `jsBuild()`, not a design decision. The existing `lib-defaults` case at line 1049 (`.withDirectory('/app/lib-defaults', ...)`, aliased in `vite.config.ts` as `@repo-lib-defaults`) is precedent for mounting a repo-root sibling into this exact container — the fix would be the same shape: mount `this.source.directory('integrations')` at `/integrations` (sibling of `/app`, matching the real repo layout) so the existing plain relative imports resolve unmodified.

I'm not making that change myself — it's shared CI infrastructure used by every branch/PR, not code this PR was asked to touch, and outside the "trivial demo plugin" scope I was given. Flagging it here since `pets` is now one more instance of it. Let me know if you'd like me to send that fix as its own PR.

---

michielbdejong:
Update: the `/browser` alias fix (previous commit) resolved the `UNRESOLVED_IMPORT`/tsconfig failures — `Main (GitHub) / CI` now gets past the frontend build entirely.

It's now failing one step later, at `pnpm -r lint` in `browser/e2e` (`oxlint ... && pnpm format-check`, "Found 26 warnings and 5 errors"). This is also pre-existing and unrelated to this PR: every flagged line already exists verbatim on this PR's base commit (`feat/plugin-model`@`3563a7ee24`, before any commit here) — e.g. the `typescript(no-explicit-any)` hits on `(window as any).store` in `plugins.spec.ts` and `installation-recovery.spec.ts:20`, and the `no-console` hits in `scripts/record-features.mjs`, are all unchanged pre-existing code (verified with `git show 3563a7ee2:browser/e2e/tests/plugins.spec.ts`). None of it is in a file this PR touches beyond appending the new Pets test near the top of `plugins.spec.ts`, which shifted these pre-existing lines' numbers but not their content.

Not fixing the e2e package's existing lint debt here — that's a separate, unrelated cleanup outside a trivial demo plugin's scope. Flagging per the same "not this PR's" reasoning as the earlier comment.

---

michielbdejong:
**CI red on `6acc3fa75` — not this PR's, looks like a resource-contention flake**

`Main (GitHub) / CI` failed in `data-browser test`: `src/oxc-react-compiler.test.ts › transforms every app TSX/JSX file without a fatal error` hit vitest's default 5000ms timeout (818/819 other tests passed). That test synchronously walks and `oxc-transform-react`-compiles every `.tsx`/`.jsx` file under `data-browser/src` in one `it()` block with no custom timeout — CPU-bound, no I/O, so a shared/loaded CI runner is enough to tip it over 5s. It predates this PR (`381a517dd`, before any commit here) and nothing in this PR touches it or its inputs.

Not spending the one re-run on this stale commit — a fresh run is already in flight for `bae16426c` (the next commit, unrelated to this test) and will exercise the same test independently. If it recurs there I'll treat it as real and either fix the root cause or give the test its own timeout budget.

---

michielbdejong:
**CI red on `740d60383` — not this PR's, pre-existing in `integrations/tooling/certify.mjs`**

`Main (GitHub) / CI` (and the `ci --host-profile hosted --playwright-mode light` dagger status) now gets past the e2e-lint, `lib` test fixtures, and `oxc-react-compiler` timeout issues fixed earlier in this thread, and reaches the JS integration-certification step for the first time — where it fails identically for **all five** integrations: `clockify: failed`, `github-issues: failed`, `mt940: failed`, `notion: failed`, `pets: failed` (`js; live not run`).

Root cause: `certify.mjs`'s `reproducible-bundle` check (`integrations/tooling/certify.mjs:154-164`, from `39f5083015` — predates any commit on this branch) shells out to `browser/node_modules/.bin/esbuild`, but `esbuild` is never declared as a dependency anywhere in the repo (`grep -rln '"esbuild"' --include=package.json .` outside `node_modules/.pnpm` returns nothing) — it's only present transitively via `vite`. Reproduced locally: `node integrations/tooling/certify.mjs --layer js --integration pets` fails with `spawnSync .../esbuild ENOENT`, and the report shows `typecheck`/`fixtures` passing while only `reproducible-bundle` fails. Since this affects `mt940`/`clockify`/`github-issues`/`notion` identically — integrations this PR never touches — it isn't specific to `pets`; it just hasn't been visible until now because an earlier-failing step always short-circuited the pipeline before this one finished and printed its summary.

Proposed fix: declare `esbuild` as an explicit devDependency (e.g. in `browser/package.json`, pinned to `0.28.1` — the version already resolved via `vite`) so pnpm reliably provisions `browser/node_modules/.bin/esbuild`. Not making that change myself — it's a workspace-root `package.json`/lockfile edit to shared tooling this PR doesn't otherwise touch, outside the "trivial demo plugin" scope I was given, same reasoning as the earlier `jsBuild()` mount-gap comment. Let me know if you'd like me to send it.

---

michielbdejong:
I still had a Claude Sonnet process that I forgot about that was undoing the fixes from GPT-6 Astra. :) This PR is now mergeable. It contains a useless "Pets" integration but that we will not want to ship in production but the good thing is that it includes Playwright tests and several fixes to make those tests pass, so I think merging this into `feat/plugin-model` could be useful now?

michielbdejong:
The [Pets E2E run for 2ca03bd](https://github.com/ontola/atomic-server/actions/runs/34349484662) passed. `pets-e2e.yml` served its purpose: it provided a focused check that the demo Pets table installs and displays after the reviewed import is applied.

I'm removing this temporary branch-specific workflow and documenting how to run focused browser journeys through Dagger in CONTRIBUTING.md. The Pets test stays in the shared full E2E suite, which runs on develop, v* tags, and feature branches opting into full E2E. This removes the extra automatic Pets-only run on this branch without removing the test.

The main pipeline for 2ca03bd was cancelled; this confirms the focused Pets journey passed, not that all CI passed.

michielbdejong:
This branch now contains the machinery for adding API imports into the an atomic-server that runs in the browser. It relies on https://localthought.io (which runs https://github.com/localthought/integration-proxy on Heroku) for API key management and CORS proxying. To use it, log in to https://localthought.io with a Google account to become a tenant and receive a tenant secret that you can paste into atomic-server and start syncing your github issues and google calendars.

The proxy exposes a catalog based on a config file that points at a subset of https://github.com/localthought/overlays and https://github.com/localthought/openapi-directory.

The imported data becomes available in its own schema using https://github.com/localthought/syncables-rs.

It is then translated using https://github.com/localthought/devonian so that there is two-way sync between a github issue tracker and an atomic issue tracker, and between a Google calendar and an atomic calendar.

We should talk about how we want to brand and market this proxy service and its surrounding ecosystem (openapi-directory, overlays, syncables, devonian), for now I branded it as Local Thought which was a domain name I still had lying around and that I can donate to Ontola if you want to use it.

Issue #1406, opened 2026-09-09
Google Calendar import fails: Unknown query parameter showDeleted


michielbdejong:
### Problem

On the deployed `feat/api-plugins` frontend, Google Calendar OAuth completes, but **Fetch and preview** fails with `Error: Resource not found. DID Resource … not found locally` before an import preview appears.

### Reproduction observed on production

1. Open `https://localthought.ai/app/integrations`.
2. Set up Google Calendar through LocalThought using an existing tenant/account.
3. Complete OAuth and return to the frontend; the callback code is removed from the URL.
4. Fetch the UTC range 2026-09-09 through 2026-10-09 (exclusive).
5. Observe the missing local DID error instead of the review dialog.

The same error occurs after creating a fresh drive and reconnecting Calendar to it. The fresh drive's default ontology is accessible and remains empty. The original remote drive also displays a missing-resource error in integration discovery; the fresh drive does not show that discovery error.

### Version and verification

Production is Heroku release v9, merge commit `08792bff96ade9cc097adae226a25741c25d0cb2` in `localthought/atomic-server:feat/heroku`, incorporating upstream `feat/api-plugins` at `0f11ca647e80ed8ca6aba0710203119acc791c6f`.

The deployment build and 31 focused LocalThought/Calendar/Devonian tests passed. This browser-state failure is not covered by those checks. No tenant secrets, OAuth credentials, or calendar contents are included here.

Expected: a bounded Calendar fetch produces a reviewable import and can be applied locally. Preserve existing data and report genuine missing dependencies rather than silently dropping them.

Root cause and regression test are under investigation.

michielbdejong:
I'll see if this is still reproducible in the latest feat/plugin-model-improvements branch

michielbdejong:
It's now failing with:
> Error: Unknown query parameter showDeleted for /calendars/{calendarId}/events

<img width="1315" height="1247" alt="Image" src="https://github.com/user-attachments/assets/a38680ef-263d-4c6c-ab20-26578033c1ad" />

michielbdejong:
[...]

PR #1407, opened 2026-09-09
Keep local integration resource reads off the server

michielbdejong:
Local integration setup queries identities in OPFS but then loads their resources through `Store.getResource`, which can contact the server while online. After a reload, a locally indexed schema that has not reached the server can therefore abort setup with a missing-resource error.

Add `Store.getLocalResource`, using the existing local JSON-AD/Loro snapshot hydration path, and use it for local installation and schema reads. Missing or unavailable local state still fails rather than allowing duplicate creation. Server-backed installations retain their existing reads.

Addresses #1406. Targets `feat/api-plugins` because the affected integration code is not yet on `develop`.

Validation:
- Regression failed before the fix: a real Store plus the local schema adapter contacted a rejecting server despite a matching local database record. It passes after the fix.
- Added checks for an indexed-but-missing installation (no duplicate creation or remote fetch) and an unavailable local database.
- 611 library tests, 820 frontend tests, and 2 MCP tests pass.
- Library and frontend TypeScript checks, formatting, and `git diff --check` pass.
- Fixed inherited integration spacing/formatting failures identified in the previous feature-branch CI run. Frontend and E2E lint now pass; all eight cleanup files have equivalent syntax trees, and the 13 Devonian tests pass.

The reproduction controls the local database and remote failure; it does not replay the affected production OPFS database. The patched Google Calendar OAuth/fetch/review flow still needs live verification after deployment. No production deployment is included in this PR.


michielbdejong:
Investigated the current PR and pushed `3ae1f0434` to its existing branch.

- The earlier run [34368249532](https://github.com/ontola/atomic-server/actions/runs/34368249532) failed on seven E2E blank-line lint errors in `devonian-issue-sync.spec.mts`. The existing `5e0074066` formatting commit had already fixed those.
- The subsequent run [34370530775](https://github.com/ontola/atomic-server/actions/runs/34370530775) was cancelled. Its aggregate Main job explicitly failed because the GitHub CI job was cancelled; the available Actions log does not establish a new code defect or explain who/what cancelled it. The separate Dagger failure status is not proof that the current code fails.
- Checked the updated base `4d7b7ad6d`, including #1408 and #1409. Integrated it into this branch and resolved the `localImportVerdict.ts` conflict while retaining the new `localImportRows` helper.
- Reproduced six additional E2E lint errors in the base's `google-calendar-import.spec.mts`; fixed them with blank lines only.

Validation on the combined branch: E2E oxlint passes; E2E formatting passes (94 files); local-schema-resource regression tests pass (3); installationResources tests pass (3); data-browser TypeScript check passes; `git diff --check` passes. Used existing local dependency installations for validation. Full Rust/build/browser CI remains to be confirmed by the fresh remote run.

No PR was merged and auto-merge was not enabled.

GitHub confirms head `3ae1f0434672ea3534d0f919d450ddcc26bbcfd9` is mergeable. Fresh CI [34374742705](https://github.com/ontola/atomic-server/actions/runs/34374742705) is pending.

Issue #1227, opened 2026-06-23
Make AtomicServer discoverable in self-hosted platforms


joepio:
Atomic-Server is a great open source all-in-one workspace that can be self-hosted. But self-hosting is still a bit of hassle, even for technical people. Many platforms offer curated repositories / lists of packages that can be discovered in their platform. This often requires some minimal prepartion work (e.g. setting up some config file with a standardized name in this repo, and adding a link in some other repo, or doing a PR for some config data + image). We already have done some of the most important prep work (CI + multi-arch builds), so the next steps are probably high-leverage.

## Which platforms are important?

- [x] Docker, obviously. We have a dockerfile, we build multi-arch images in the CI. 
- [x] Docker Compose. Not really needed, but we have it. #949 
- [ ] Helm Chart. 
- [ ] HomeAssistant #1138 See [this repo](https://github.com/ontola/hassio-atomic-server). Not yet in HA core - so not very discoverable still.
- [x] ArtifactHub [![Artifact Hub](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/atomic-server)](https://artifacthub.io/packages/search?repo=atomic-server)
- [ ] [Unraid](https://forums.unraid.net/topic/38582-plug-in-community-applications/)
- [ ] [LinuxServer.io](https://www.linuxserver.io/)
- [ ] [TrueNas Apps](https://github.com/truenas/apps)
- [x] [Awesome-Selfhosted](https://github.com/awesome-selfhosted/awesome-selfhosted)
- [x] [JamStack](https://jamstack.org/headless-cms/atomic-server/) (for headless CMS tools)
- [ ] [Vultr Marketplace](https://www.vultr.com/marketplace/become-a-verified-vendor/) #157
- [ ] AWS Marketplace 
- [ ] [CasaOS AppStore](https://github.com/IceWhaleTech/CasaOS-AppStore/blob/main/CONTRIBUTING.md)
- [ ] DigitalOcean Marketplace ([make a PR here](https://github.com/digitalocean/droplet-1-clicks/tree/master))

joepio:
[...]

PR #1422, opened 2026-09-10
docs: add Easypanel deployment option

Ahson-Shaikh:
Adds Easypanel as a 5th setup option alongside Docker/binary/cargo/source. Atomic Server has an official one-click template on Easypanel (https://easypanel.io/templates/atomic-server). Note: this also corrects the GitHub repo link tracked for this tool — the sheet previously listed joepmeneer/atomic-server, which 404s; the active repo is ontola/atomic-server (mirrored at atomicdata-dev/atomic-server).

### DIFF
--- a/docs/src/atomicserver/installation.md
+++ b/docs/src/atomicserver/installation.md
@@ -8,6 +8,7 @@ You can run AtomicServer in different ways:
 2. From a published [binary](https://github.com/atomicdata-dev/atomic-server/releases)
 3. Using [Cargo](https://doc.rust-lang.org/cargo/getting-started/installation.html) from crates.io: `cargo install atomic-server`
 4. Manually from source
+5. Using [Easypanel](https://easypanel.io/), a server control panel with an [official one-click template](https://easypanel.io/templates/atomic-server)
 
 If you want to run AtomicServer locally as a developer / contributor, check out [the Contributors guide](https://github.com/atomicdata-dev/atomic-server/blob/develop/CONTRIBUTING.md).
 


joepio:
Hey Ahson,

Thanks for the PR. I could linkt people to this, but it seems to use a quite outdated version (0.38), can I edit this somewhere?

Ahson-Shaikh:
Hi @joepio - I can update that and loop that here for your reference.

Ahson-Shaikh:
Hi @joepio - thanks for flagging this. Dug in a bit: our template was pinned to `joepmeneer/atomic-server:v0.38.0`, which turned out to be an abandoned fork. The active image is `ghcr.io/ontola/atomic-server`, but it looks like your CI only ever pushes a rolling `develop` tag there - no versioned tags (v0.40.3 etc.) exist on the registry. So I've pinned our template to the current `develop` digest by SHA instead (our repo's convention requires an exact pin, never a floating tag), which is ahead of v0.40.3 and picks up the SSRF/invite-reuse/path-traversal/XSS fixes from that release. PR: https://github.com/easypanel-io/templates/pull/1578

To answer your question - the template config lives at https://github.com/easypanel-io/templates/tree/main/templates/atomic-server (meta.yaml has the image default). You're welcome to open a PR there directly any time you cut a release, or just ping this thread and I'll bump it. Also - if you ever start pushing actual version tags to ghcr.io/ontola/atomic-server, that'd let us (and anyone else) pin to a real release instead of a digest, which would be more sustainable long-term.

joepio:
Thanks for digging in. We're in the middle of a big new release, which will update `latest` on both the dockerhub and ghcr ones, not just `develop`. 

Also thanks for the link to template, I'll refer to that one from my own issue that keeps track of external deploy options (#1227)

Ahson-Shaikh:
Sounds great, thanks for the update @joepio - I'll keep an eye out for the new release and bump the template's digest pin once `latest` is properly tagged on your end. Appreciate you linking to it from your own tracking issue too.

PR #1426, opened 2026-09-11
Add Codex chat app and local worker

michielbdejong:
Adds a Codex integration to Atomic's catalog with a source-as-data chat view and local worker. Users install it from Integrations, download an app-scoped worker setup, select a local workspace, and run the worker with their Codex CLI sign-in.

Sending a message persists a queued turn in Atomic first. The worker reads it, calls its dedicated `codex app-server --stdio` process, and saves replies and status back to Atomic. Conversations support continuation, streamed text/output, Stop, and approve-once/decline for command/file requests. A local journal prevents automatic replay after uncertain execution.

Full-browser testing exposed native form submission being blocked by the iframe sandbox and cached reads hiding worker updates. The composer now uses an explicit button handler, and the host refreshes view reads. Existing iframe permissions and RPC operations are unchanged.

Validation:
- Eight worker/protocol tests and twelve host tests pass.
- Sandboxed Chromium regression passes for sending, replies, switching, approvals, Stop, escaping and mobile layout.
- Frontend and example TypeScript checks and example build pass.
- Full Data Browser installation and downloaded worker setup verified against an isolated AtomicServer.
- A UI-submitted prompt was independently read from Atomic as queued while the worker was stopped. The real worker then saved a response; a second turn retained context. Both replies rendered inside Atomic's actual app view.

The screenshots below are from the full Data Browser and show a real saved Codex conversation. Live checks used an existing local AtomicServer binary; Rust was not rebuilt. Live command/file escalation is not certified. Attachments, full diff review, desktop-history import and multi-worker coordination remain outside this example's scope. Setup and recovery details are in `plugin-examples/codex-chat/README.md`.

### Integration catalog

Codex in Atomic's integration catalog

### Chat inside Atomic

Real Codex conversation hosted inside Atomic


michielbdejong:
I ran into a dilemma with this: if we do it as a plugin then it's separate from the already-built-in AI assistent with vercel / openrouter / ollama. But if we want it to nicely go into the AI assistent then it cannot be a plugin, and we would need to ship it in the main code, which is not defendable for such a niche feature. So parking this for now and working on other things first.

PR #1484, opened 2026-09-14
Add read-only HTML source view for documents

michielbdejong:
Documents now offer **View as HTML** in the resource menu, including Data View. It opens a read-only, copyable HTML snapshot of the actual Loro document body, so an empty legacy `document-content` property no longer prevents inspection. WYSIWYG remains the editing interface.

The viewer uses TipTap’s existing document serializers and adds line breaks between top-level blocks, preserving inline and code whitespace. It is available to readers, loads lazily, and distinguishes loading, missing state, pending edits and conversion errors from an empty document. Inspection uses detached state and avoids committing pending edits. Existing AI/diff readers retain their behavior.

Validation: 27 targeted unit tests and the focused Chromium HTML-source E2E (1 passed, strict diagnostics), production Vite build and repository-wide pre-commit lint/format checks. Typechecking reports the pre-existing `src/hooks/useFile.test.tsx:36` TS2352 cast error. UI translations were extracted with Vite and the task-related catalog changes reviewed. Local browser acceptance uses an isolated backend, database and production frontend.


michielbdejong:
Screenshot 2026-09-14 at 14 07 17
Click "View as HTML"

Screenshot 2026-09-14 at 14 07 05

michielbdejong:
HTML export fidelity notes:

HTML is not guaranteed to preserve everything in TipTap: each extension decides what to serialize and can omit attributes ([TipTap documentation](https://tiptap.dev/docs/editor/extensions/custom-extensions/extend-existing#attributes)).

In this implementation:

- Resource embeds become links containing the resource's subject and title. Their interactive contents are not expanded.
- Notes retain their contents inside `<note-block>`, but their appearance depends on the app's styling.
- Image captions and float settings remain as attributes, but ordinary browsers will not display those attributes as captions or positioning automatically.
- Standard text, headings, lists, tables, and formatting have HTML representations.

Some information therefore remains visible in the HTML source without producing the same visual result when rendered. TipTap JSON is the more precise representation of document structure. Loro's collaboration history is separate and appears in neither export.

michielbdejong:
You can also click the word 'New' to create a resource, give it the https://atomicdata.dev/classes/PlainText class and then edit the description.

You can also select a word in the WYSIWYG editor to see the underlying TipTap:
Screenshot 2026-09-14 at 14 23 42
Screenshot 2026-09-14 at 14 23 53


Since the 'View as HTML' is a read-only snapshot and it's a lossy translation from the TipTap, and we have these two other features for escaping the WYSIWYG, I think we can shelf this proposal.

Issue #434, opened 2022-06-13
Atomizer


joepio:
See https://github.com/tauri-apps/tauri-plugin-log

## What do we need to build?

- Importer for json-ad #390 
- Recognize file types
- Markdown file importer

joepio:
Relates to #1134

PR #1485, opened 2026-09-14
Convert uploaded Markdown and text files to documents

michielbdejong:
Uploaded Markdown and plain-text files now offer **Convert to document** beside Download for users with write access. Conversion opens the editable document at the same URL, preserves location, permissions and the original blob, and imports Markdown formatting or literal text with line breaks.

The conversion prepares content against the file's existing Loro history, rechecks permissions and concurrent conversion, and offers Retry if saving fails. A related partial test-fixture cast was corrected to unblock frontend typechecking.

Validation: 12 focused conversion/file-preview tests, frontend typecheck, targeted lint/format checks, and a live Chromium upload → Convert to document → edit → reload check against an isolated real backend. Filename-only Markdown uploads, plain-text fidelity, metadata preservation and error paths were also checked. Public file documentation and the test-coverage map are updated.


joepio:
My instinct is to merge these efforts with atomizer #434 but maybe for this one it makes more sense to just convert in-browser, as the browser has the JS logic to create documents.

michielbdejong:
Makes sense! Better to have a generic process for what to do with an incoming (uploaded) file. Closing this in favour of #434.

PR #1506, opened 2026-09-15
Dedupe ontology property shortnames when creating table columns

### LINKED ISSUE #1504: Dedupe shortnames when creating table columns to prevent ontology ambiguity
## Background

While fixing #1503 (crash on the workspace Connections dialog, and silently broken GitHub issue sync), we found the underlying data problem that triggers it: a drive's ontology can end up with **two different property resources sharing the same `core:shortname`** (e.g. `status`), because column creation never checks for an existing property with that shortname before minting a new one.

This doesn't live in the GitHub issues (devonian) plugin itself — that plugin (and its bridge/adapter code) correctly reuses the shared, fixed `taskSchema.properties.status` subject (`https://atomicdata.dev/task/v1/status`) rather than creating its own. The duplication comes from **generic table/column creation**, which any "New Table" or "New Column" flow goes through — including the devonian demo's own local tracker setup, which is one of several callers.

## Root cause

- `createPropertyOnClass` and `createSelectPropertyOnClass` in [`browser/data-browser/src/chunks/TablePage/Kanban/createSelectProperty.ts`](browser/data-browser/src/chunks/TablePage/Kanban/createSelectProperty.ts) always mint a brand-new property resource, setting its shortname via `stringToSlug(opts.name)`, with no check for whether a property with that shortname already exists elsewhere in the drive's ontology.
- `attachPropertiesToClass` (same file) then pushes the new property straight into the drive's shared default-ontology `core:properties` list.
- Result: creating a second column named "Status" (in any table, from any flow — the New Table dialog, a template, the AI assistant's `create_table` tool, or `browser/data-browser/src/chunks/DevonianDemo/demo.mjs`'s local Issue Tracker setup) silently creates a second, distinct property resource that also slugs to shortname `status`. Nothing fails at creation time — the ontology just quietly becomes inconsistent.
- That inconsistency later breaks any code that resolves properties/classes by shortname across the whole ontology (`browser/lib/src/plugin-schema.ts`'s `byShortname`, before #1503's fix, threw on this for entirely unrelated lookups; `browser/data-browser/src/chunks/AI/jsonAdCompact.ts`'s shortname resolution for the AI assistant has the same kind of "ambiguous shortname" check and would hit the same problem for any assistant operation that references the colliding shortname directly).

## Suggested fix

Before creating a new property in `createPropertyOnClass` / `createSelectPropertyOnClass`, check whether a property with the same shortname already exists in the target ontology:
- If one exists and is compatible (same datatype, and for `select` columns the same `dataBrowser.classes.selectProperty` shape), reuse it — attach the existing property to the new row class instead of minting a duplicate (similar to the existing `column.propertySubject` reuse path in `createColumnOnClass` in `createTableFromSpec.ts`).
- If one exists but is incompatible (different datatype/shape), either mint the new property with a disambiguated shortname (e.g. `status-2`) or surface a clear error/choice to the caller — don't silently create a second resource with the same shortname.

## Scope / out of scope

- This is a shared table/column-creation concern, not something specific to the GitHub issues integration or the devonian plugin — please don't scope the fix to `integrations/github-issues/**`.
- Out of scope: changing how `browser/lib/src/plugin-schema.ts` or `browser/data-browser/src/chunks/AI/jsonAdCompact.ts` handle ambiguity — #1503 already scoped `plugin-schema.ts`'s check to the shortnames a given lookup actually cares about, which is enough to stop unrelated collisions from crashing unrelated features. The remaining issue is genuinely-relevant collisions (an actual attempt to read/write the colliding shortname), which this issue's fix should prevent from occurring in the first place.

## Test plan for whoever picks this up

- A regression test creating two tables, each with their own "Status" select column, on the same drive — assert the second call reuses the first's property subject rather than creating a duplicate with the same shortname (or, if incompatible, gets a clearly distinct shortname).
- Run `pnpm vitest run` in `browser/data-browser` and `browser/lib`, and `pnpm typecheck` in `browser/data-browser`.

michielbdejong:
## Summary
- Fixes #1504: creating a second column named e.g. "Status" in any table (New Table dialog, a template, the AI assistant's `create_table` tool, or the devonian demo's local tracker setup) silently minted a second, distinct property resource that also slugged to shortname `status`, leaving the drive's ontology ambiguous.
- `createPropertyOnClass`/`createSelectPropertyOnClass` (`browser/data-browser/src/chunks/TablePage/Kanban/createSelectProperty.ts`) now look up the target ontology's existing properties by shortname before minting a new one: a compatible match (same datatype, and for select columns the same tag-backed shape) is reused instead of duplicated; an incompatible one gets a disambiguated shortname (`status-2`).
- Fixed two smaller bugs surfaced while testing the new reuse path:
  - `attachPropertiesToClass` could re-add an already-registered property to the ontology's `core:properties` list, producing a duplicate entry.
  - The reuse path matched a caller's requested select option names against `core:name`, but tags are only ever given a `core:shortname` — so it always failed to find any option.

## Test plan
- [x] Added `createSelectProperty.test.ts` covering: select-property reuse, plain-property reuse, shortname disambiguation on an incompatible collision, rejecting a requested option the reused property doesn't have, and no-op behavior when there's no shared ontology to dedupe against.
- [x] `pnpm vitest run` in `browser/data-browser` (1006 tests passing).
- [x] `pnpm typecheck` in `browser/data-browser`.
- [x] `pnpm oxlint`/`oxfmt` on the changed files.


michielbdejong:
Closing as redundant — this PR's commit (`0ee9b2138`) has been cherry-picked directly onto `feat/plugin-model-improvements` in [0be7b2314](https://github.com/ontola/atomic-server/commit/0be7b2314), so the fix is already on the base branch this PR targets. No further action needed here.

---

PR #1509, opened 2026-09-15
fix(wasm): resolve $ref query parameters before override validation

michielbdejong:
## Summary
- `apply_query_overrides` (wasm/src/calendar_import.rs) validated a catalog's declared query parameters against the raw, unresolved OpenAPI document — `$ref` expansion only happens later, in `load_open_api_document`. Real-world catalogs (Google Calendar's included) commonly declare a shared query parameter once under `components.parameters` and reference it via `$ref` on each operation instead of inlining `{name, in, schema}`. Any such parameter was reported as "unknown" even though the catalog does declare it, producing `Unknown query parameter showDeleted for /calendars/{calendarId}/events` when installing the Google Calendar (Devonian) plugin.
- Fix: dereference a single `$ref` hop against the document before checking whether a parameter is declared. Added a regression test (`accepts_a_parameter_declared_via_a_shared_ref`) using a `$ref`-shaped fixture, matching how a real spec (not the existing inlined mocks) declares parameters.
- Also fixes 4 pre-existing clippy failures in `wasm/src/lib.rs` (a deprecated async-constructor lint, `#[allow(deprecated)]`'d with a comment explaining why; and 3 redundant-borrow-in-`format!` lints) that were blocking the workspace pre-commit hook independently of this change.

## Note on pre-commit
This commit was made with `--no-verify` at the requester's explicit direction. `wasm/tests/vault.rs` currently has 4 pre-existing argument-count mismatches against `vault_export`/`vault_import` that also fail the hook's `--all-targets` clippy check. That's unrelated to this fix and touches vault-crypto code, so it wasn't guess-fixed here — flagging it so it doesn't get lost:
```
cargo clippy -p atomic-wasm --target wasm32-unknown-unknown --no-deps --all-targets --no-default-features -- -D warnings
```

## Test plan
- [x] `cargo check -p atomic-wasm --target wasm32-unknown-unknown` compiles cleanly
- [x] `cargo fmt -p atomic-wasm -- --check wasm/src/calendar_import.rs` passes
- [x] Verified the fix against a standalone extraction of `apply_query_overrides` (the crate is `wasm32`-only, so its `#[cfg(test)]` unit tests don't run via plain `cargo test`): reproduced the exact reported error with a `$ref`-declared parameter, confirmed it's fixed, confirmed the existing inline-parameter case still passes
- [ ] `wasm/tests/vault.rs` still fails to compile (pre-existing, unrelated — see above)


michielbdejong:
Blocked on ontola/atomic-server#1507.

PR #1524, opened 2026-09-16
docs(planning): audit the data-browser UI and propose a token-first refactor

joepio:
Measures the design-system drift in `browser/data-browser/src` rather than
describing it: 56 distinct font-size values across 334 declarations, 53 gap
values, 108 padding values, 39 box-shadows and 37 durations, against a theme
object used 263 times for spacing where 2216 literals are written instead.

Proposes fixing the token layer before any restyle: CSS custom properties on
`:root` (OKLCH ramps, a 7-step type scale, one elevation ladder), a per-property
codemod whose real deliverable is the lint rule that keeps the drift out, then
surface consolidation following the `cardSurface` / `floatingSurface` precedent.
Chrome and zero-runtime-CSS work come after, and a visual redesign is explicitly
not the first move.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01BgT3HPa7EmbpVayvxDXJW8


joepio:
@Polleps do you agree with this change?

PR #1307, opened 2026-08-25
feat(apps): custom UIs, schemas and logic on Atomic data

joepio:
## Why: apps inside Atomic

This PR implements the app foundation proposed in **#1304 — “Apps inside Atomic: vibecode entire UIs + schemas + logic”**, including app-scoped identities related to #1275.

The goal is to let someone ask the Atomic assistant to build an app that works with their data. Atomic supplies persistence, identity, access control and synchronization; the app supplies its schema, interface and behavior. Sharing the app and its data uses Atomic resources and permissions rather than requiring a separately hosted application.

For example: “Build a project tracker” can produce a custom interface over Atomic records. Those records remain accessible in native tables and boards. A GitHub connection can synchronize issues into the tracker, and a separate automation can act when a new issue arrives.

This branch also develops the integration and automation infrastructure needed for that model. **The foundation is implemented, but migration to one complete package/installation lifecycle is not finished.**

## The app model

**An app brings together data, views and optional behavior.** A custom `App` resource is a parent whose children can include its ontology, executable view and handlers. It can reuse an existing class or data table instead of inventing another schema. Native tables and folders can already serve as workspaces; they do not need an extra App wrapper.

| Concept | Meaning |
| --- | --- |
| **App / workspace** | The work users open, share and organize: for example, a project tracker. |
| **View** | A native table/board or a custom JavaScript interface over that data. |
| **Integration / connection** | A reusable connector, and its configured instance pointing to an account/source and destination. |
| **Automation** | An independently editable JavaScript script that can use Atomic data and connection actions. |
| **Package / release** | A distributable definition, and the exact executable revision used for approved work. |
| **Installation** | The local identity, configuration and permissions under which code operates. |

An app needs neither an integration nor an automation. A connection can synchronize without an automation. An automation can use several authorized connections. “Integration” describes a capability, not a reason to build another permission system, scheduler or importer. Provider-specific setup adapters still exist and are a migration gap, not the intended extension mechanism.

## What apps can do in this branch

- Render a custom interface in an isolated iframe, read and edit scoped Atomic resources, and subscribe to updates. Native views remain available for the same data.
- Define local schemas or bind to existing schema identities. App-owned records default to nested containment; explicit existing-table destinations are preserved.
- Run JavaScript that proposes imports, changes or validation problems. Automations can run manually or through the implemented event/schedule paths.
- Use declared external operations and named integration actions. The UI, assistant, JavaScript automations and MCP adapter share the host action API.
[...]

PR #1527, opened 2026-09-16
Publish integration catalog and make private-home recovery optional

michielbdejong:
Adds a published integration catalog and serves integration bundles as static server assets. Catalog visibility and plugin toggles control discovery, with preferences saved in the background.

Release and E2E builds now include the catalog and bundles. Bundle loaders use the connected Atomic server, isolate their caches by server URL, and retry after failed network, HTTP or response-body reads.

Also resolves the inherited sign-in regressions exposed by the current smoke suite: an account without recoverable data receives a writable derived private home, including when restoring a persisted session through a direct link. Connecting another device or restoring a backup is an optional nudge. Existing data is read/recovered first; unrelated workspace links are never synthesized. Sync no longer claims that an unreadable workspace is cached locally or known to live on another device. Explicit sign-out and device lock take precedence over the account-settings unlock guard.

Validation:
- 1,192 frontend unit tests; frontend typecheck.
- Four Dagger build-input regression tests.
- Five Chromium regressions pass: new and legacy secrets, persisted-session direct links, document creation/reload persistence, truthful missing-workspace status, and sign-out returning to welcome after reload.
- Translation catalogs settled through Vite and reviewed.

The local browser check uses isolated services and existing local server/WASM binaries. CI builds and tests this PR's full source. Flutter's separate sign-in flow is unchanged.


michielbdejong:
\`integrations/catalog.json\` (added on this branch, forked from #1307's \`feat/plugin-model\` at plugin model improvements (#1482)) is a temporary gate: it just toggles \`experimental\` visibility for the hardcoded \`BundledIntegration[]\` list in \`IntegrationDiscovery.tsx\` (GitHub, Google Calendar, MT940, Clockify, Notion).

#1307 already introduces the real long-term mechanism for this: the package/release model behind \`/plugin-catalog\` (\`server/src/handlers/plugin_release.rs\`, \`CatalogEntry\` in \`lib/src/db/plugin_release.rs\`, rendered as "Community plugins" in \`IntegrationStore.tsx\`). Its own remaining-work list calls the bundled provider-specific setup adapters "a migration gap, not the intended extension mechanism" and has "removal of bundled provider installer adapters" as unfinished work.

So the plan is: these bundled integrations get published through \`/plugin-catalog\` like any other package/release, and \`catalog.json\` + the hardcoded \`BundledIntegration[]\` gating go away once that migration lands.

michielbdejong:
Yeah and for now I want to see a clean shippable atomic-server UI that has just the Pets plugin as an experimental bundled plugin available. I will then do per-plugin feature branches to get each of them to a quality level where we think they are shippable as bundled plugins. And bundled here means they are in this monorepo but not built in the server build or in the data-browser build; their are loaded into the running server over https. For now, they *are* shipped in atomic-server, but as assets that are hosted as same-host statics, and they are still tied to the server's release cycle. In the future we can keep them in this monorepo or not, but stop shipping them as static assets, and instead host them somewhere. That way we can make them available to all running instances, untied from the server's release cycle. And then the catalog.json also becomes the planned database version.

michielbdejong:
I reduced this to the smaller scope so it becomes reviewable. This PR puts the bundled plugins into the server as static assets that the data browser fetches from its own (local)host over http. There is a catalog.json where we can disable plugins that are not ready for prime time yet. Only the Pets demo plugin is enabled by default.

michielbdejong:
Ready for review
Screenshot 2026-09-22 at 14 08 54

michielbdejong:
Not sure why the main pipeline was not triggered by that last commit, [triggered it manually](http://github.com/ontola/atomic-server/actions/runs/35725618853)

PR #1531, opened 2026-09-16
Harden save durability, verify sync delivery, and add private diagnostics

Polleps:
Acknowledged saves and completed syncs must survive a process crash. This PR fixes paths that reported success before durable storage or complete delivery, adds regressions at those boundaries, and makes intermittent failures easier to investigate through private, locally recorded diagnostics.

### Durable saves and verified synchronization

- Flush server commits and imported sync state before acknowledging success. Flush failures reject completion, including matching-hash probes and retries of already-readable updates.
- Distinguish a child queued behind an unsaved parent from a durable save with the `SaveResult` value `queued`. Keep local-only commits available after persistence failures and expose save errors with a retry action.
- Preserve stored CRDT history when restoring resources, processing incoming updates, and assembling collection results. Reconciliation exports the union of durable and mounted state so a stale in-memory resource cannot omit acknowledged offline edits.
- Wait for outgoing chunk acknowledgements and incoming local persistence before completing browser reconciliation. Verify that the receiver's version vectors cover the sent updates, retry missing updates within a fixed bound, and retain local data when verification fails.
- Apply equivalent verification to Iroh synchronization before reporting success. Preserve interleaved live frames and signed history envelopes on retries, and serialize Node database operations to avoid observing partially written resource state.

### Private diagnostics for feedback and AI investigation

- Record save, persistence, outbox, connection, and reconciliation events locally by default. Use rolling IndexedDB storage bounded to 500 events across 20 sessions from the last ten minutes, with writes batched every two seconds.
- Collect allowlisted metadata and anonymous resource aliases rather than document contents, filenames, URLs, credentials, or raw error messages. Provide controls to stop and clear recording, with memory-only fallback when persistent storage is unavailable.
- Upload diagnostics only when explicitly included in feedback. Put **Include diagnostic data** between feedback and email, with preview/download controls in a collapsed Diagnostics section. Sanitize feedback events so they do not inherit unrelated Sentry breadcrumbs or user context.
- Export schema-versioned reports with build information, operation/sequence identifiers, event definitions, and completeness limits. Add `browser/DIAGNOSTICS.md` and an `AGENTS.md` pointer for evidence-based AI triage.

### Recovery coverage and developer checks

- Add browser and server process-crash tests, a two-process Iroh receiver-kill test, lost-acknowledgement and interrupted-sync integration tests, and seeded failure schedules covering delayed/dropped/duplicate delivery, failed flushes, reloads, deletion, and stale replay.
- Preserve collection membership while server-only resource details are loading, with a regression that also verifies completed re-parenting still removes members.
- Keep crash harnesses independent of the library's background flush so periodic persistence cannot hide an early acknowledgement.
- Add a Vault restore drill that checks restored metadata against external S3 files, including missing/corrupt-file controls. Vault backups do not contain the attachment bytes.
- Add advisory React Compiler hooks and a manual verification command, fix the affected feedback controls, and update stability planning, protocol documentation, and the coverage map.

## Validation

Validated locally after rebasing onto `develop`:

- Workspace Rust tests: **1,078 passed**, 31 ignored, with retries disabled.
- Client unit tests: **949 passed**; app unit tests: **1,147 passed**; Svelte: **3 passed**; workspace Node test suites passed.
- Real-server integration tests: **8 passed**, one existing skipped case.
- Most recent full Chromium run with hosting enabled: **285 passed**, 10 skipped, one server-only reload failure. Investigation confirmed the child was durable on the server; loading placeholders could wrongly remove query members during hydration. A library regression reproduced it and now passes. Both server-only reload specs pass ten repetitions each (**20/20**, no retries) with the fix. A fresh complete run is in progress for `e506339b9`.
- Both inline-editing specs passed five repetitions each (**10/10**) after their fixture readiness fix and passed in the most recent full run.
- Opt-in publishing/media and catalog checks: **10 passed**.
- Workspace TypeScript checks, workspace build, browser lint/format, Rust formatting, Clippy, and CI orchestration regression tests passed.
- Full GitHub CI for `e506339b9` was requested: https://github.com/ontola/atomic-server/actions/runs/35634185948. Merge remains gated on a clean complete local E2E run and any enforced GitHub protections.

## Limits and follow-up

These tests establish process-crash recovery, not physical power-loss guarantees. Older peers do not gain durable acknowledgements from version-vector verification alone, and completed resource sync does not establish that attachment downloads have finished. Diagnostic batching can lose the last batch on a crash; reports explicitly describe that limit. Deployed Sentry receipt, attachment, source-map symbolication, and backend acceptance remain unverified pending access and deployment of the intended build.

## Checklist

- [ ] Add release changelog entry linking the PR and describing the `SaveResult` API change.
- [x] Add or update tests.
- [x] Update documentation and stability planning.


joepio:
Triaging the Sentry backlog, one issue turns out to come from this branch: [`ATOMIC-BROWSER-S`](https://ontola.sentry.io/issues/ATOMIC-BROWSER-S), first seen 17 September on `atomic-data-browser@0.41.0-beta.7+980ec5d61`.

```
InvalidStateError: Failed to execute 'abort' on 'IDBTransaction': The transaction has finished.
  at src/helpers/diagnostic-storage.ts
```

Sentry tags it `mechanism: auto.browser.browserapierrors.setTimeout` and `handled: no`, which points at the watchdog in `IndexedDiagnosticStorage.update`:

```ts
const tx = db.transaction('state', 'readwrite');
const timeout = setTimeout(() => tx.abort(), 3000);
```

`oncomplete`, `onabort` and `onerror` all clear the timer, so the common paths are covered. But `tx.abort()` throws if the transaction has already finished, and here it is called bare inside a timer callback, so there is at least one ordering where the timer still runs against a finished transaction. I have not reproduced it, so I can't tell you which; the event is the evidence that it happens.

Worth guarding regardless, because of where the throw lands. It escapes the timer rather than the promise executor, so `reject` is never reached and the `update()` promise stays pending forever, on top of the unhandled error. Something like:

```ts
const timeout = setTimeout(() => {
  try {
    tx.abort();
  } catch {
    reject(new Error('diagnostic storage unavailable'));
  }
}, 3000);
```

The `catch { tx.abort(); }` in `request.onsuccess` runs inside an active transaction, so that one looks fine as it stands.

Unrelated to the diff: CI here is red with `start engine: new client: context deadline exceeded`, which is the Dagger engine on Mancave rather than anything in the branch.

---

PR #1539, opened 2026-09-17
Reconcilable host-to-Drive mappings, for hosted vanity subdomains

joepio:
Paired with **ontola/atomic-saas** on the same branch name (`claude/custom-subdomains-cloud-ioi2ex`), which adds the alias registry, plan gating and the node-side reconcile. Per `AGENTS.md`, merge this side first.

`Tree::DriveMapping` already routes a hostname to a Drive, but the only way in was `/bind-drive`, which needs an agent with write access on the Drive. A control plane that deliberately does not hold user signing keys cannot use it, so hosted vanity subdomains (`acme.atomicserver.eu`) had no way to be installed. This adds the pieces the control plane needs, and closes a leak found on the way.

### What changed

- **`Db::sync_drive_mappings`** reconciles the mappings a control plane owns against a desired set: adds, repoints and removes, idempotently, so a managed node can call it on every policy poll and a released name stops resolving with nobody touching the box. Removal is scoped to the hosts a previous run installed, tracked in `Tree::PluginMeta` — treating the input as authoritative for the whole table would have deleted the `localhost` / `127.0.0.1` entries every store gets at setup, and any binding made by hand, on the first run.
- **`Db::list_drive_mappings`** and **`Db::managed_alias_hosts`** expose the table, which had no read path beyond a single-host lookup.
- **Mapping keys are normalized.** Hostnames are case-insensitive, but the `Host` header arrives in whatever case the client sent, so `ACME.example.com` missed the mapping `/bind-drive` had stored lowercased and fell through to the store root.
- **A host bound to a Drive now serves that Drive or 404s.** `map_request_subject` fell through to the store root when the bound Drive was missing — not synced yet, or migrated to another node — which on a multi-tenant node answers one tenant's hostname with another namespace's content. This is also the property that lets the control plane authorize a certificate on reservation rather than only after a node confirms: the worst case in that window is a 404, not a leak.
- **`--served-domain-suffix`** makes a server answer for `*.<suffix>` without the rest of `--base-domain`. Nodes need the request origin to follow the hostname the visitor used, or served subjects localize to the node's own name and a vanity host hands out links to a different hostname than the one the visitor typed. `--base-domain` does that too, but it *also* becomes the store's base domain, changing how subjects are normalized and migrated — not something to switch on for a box that already holds customer data. This is the conflation `config.rs`'s own note on `ATOMIC_BASE_DOMAIN` flags.

### Related Issues

No issue; this is the `atomic-server` half of hosted vanity subdomains, designed in `atomic-saas`'s `planning/VANITY_SUBDOMAINS.md`.

### Checklist

- [x] Add changelog entry linking to issue, describe API changes — `CHANGELOG.md` covers the new `atomic_lib` API, the 404 behaviour change and the new flag
- [x] Add or update tests if needed — `db::drive_mapping_tests` (add/repoint/remove, idempotence, ownership scoping, case-insensitivity, empty inputs), `db::resolver_tests::a_bound_host_whose_drive_is_missing_does_not_serve_the_store_root`, `context::tests::a_served_domain_suffix_accepts_tenants_without_a_base_domain`
- [x] Update docs if needed — `TESTING_COVERAGE.md` gains a section for host-to-Drive routing, including what is still uncovered

### Verification

- `cargo test -p atomic_lib --features db-redb --lib` — 550 passed, 0 failed
- `cargo test -p atomic-server --lib` — 86 passed, 0 failed
- `cargo clippy -p atomic_lib -p atomic-server --all-targets --locked` — clean
- `cargo fmt --check` — clean

Not covered: no test drives a real HTTP request against a vanity host end to end. The reconcile and the resolver are tested separately, and joining them needs the representative two-service environment.

https://claude.ai/code/session_01WJmx2z4BWAr4gNvTvg48D4

---
_Generated by [Claude Code](https://claude.ai/code/session_01WJmx2z4BWAr4gNvTvg48D4)_


joepio:
`Main (Mancave) / CI` is red on `90f6a45`, and the failure is not this PR's.

**What is failing:** one data-browser vitest, `src/views/File/convertFileToDocument.test.ts > fileContentsToTiptapJson > uses the collaborative Markdown schema so Markdown formatting becomes document nodes` — `Error: Test timed out in 5000ms.` Everything else passes: **1086 of 1087 tests, 134 of 135 files.**

**Why it is not this PR's:**

- This PR's diff is `lib/src/db.rs`, `server/src/config.rs`, `server/src/context.rs`, `CHANGELOG.md` and `TESTING_COVERAGE.md`. It contains no TypeScript and nothing under `browser/`.
- The same single test failed identically on this branch's earlier head `f00feb2`, **before** the 206-commit `develop` merge, and again on `90f6a45` after it. Two trees 206 commits apart, same test, same 5 s timeout.
- It is a timeout, not an assertion failure. The run reports `transform 172.05s, import 282.84s` against an 81 s wall — the runner is heavily oversubscribed. That is also consistent with the rest of today's runner trouble on this box (#1530 could not start its Dagger engine at all; these jobs sat queued from 21:35 to 00:10).

**No fix for it exists to port**, so rather than widening this PR with a change to an unrelated test file, here is the proposed patch:

```diff
-  it('uses the collaborative Markdown schema so Markdown formatting becomes document nodes', async () => {
+  it('uses the collaborative Markdown schema so Markdown formatting becomes document nodes', async () => {
     ...
-  });
+  }, 30_000);
```

That test is the only `async` one in the file, and its cost is dominated by a one-off cold import chain rather than by the assertion: it constructs a `Store`, pulls in the collaborative Markdown schema and TipTap extensions, and loads Loro's Node WASM through `vi.mock('loro-crdt', () => import('loro-crdt/nodejs'))`. Vitest's default `testTimeout` is 5000 ms. Giving this one test an explicit timeout keeps it running and keeps every assertion — it does not skip, disable or quarantine anything.

I have not applied it here, because it belongs in its own change against `develop` rather than in a PR about host-to-Drive routing. Happy to open that PR if wanted.

Local verification of what this PR actually changes, against the merged `develop`: `atomic_lib` 601 passed / 0 failed, `context::tests` 5/5 including the new `a_served_domain_suffix_accepts_tenants_without_a_base_domain`, `cargo fmt --check` clean, and `scripts/check-rust-alignment.py` reports Server and SaaS aligned.

---

PR #1548, opened 2026-09-18
Make the Clippy pre-commit hook incremental

### LINKED ISSUE #1533: Pre-commit hook rebuilds the whole workspace
Pre-commit hook rebuilds the whole workspace on every commit (~2.5 min cold, slow warm)

`scripts/pre-commit.mjs` takes ~2.5 minutes on a cold build and is not
cheap warm either. This bites anyone running Claude Code in a fresh
worktree or in a cloud sandbox, where commits are frequent and `target/`
starts empty.

## Why it is slow

The hook `git checkout-index`es the staged tree into a fresh `mkdtemp`
directory and runs

    cargo clippy --workspace --exclude atomic-server-tauri --no-deps
      --all-targets --no-default-features --features light -- -D warnings

there, with `CARGO_TARGET_DIR` pointed at the repo's `target/`.

1. **Every workspace crate recompiles on every commit, even warm.**
   Cargo fingerprints by mtime. `checkout-index` writes every file with
   the current time, so all workspace members (and their `--all-targets`
   test targets) look modified on every run. Only crates.io deps are
   reused.
2. **Feature mismatch with the dev loop.** `--no-default-features
   --features light` is a different feature graph from what
   `cargo build` / `cargo test` use, so feature-sensitive deps are
   compiled twice into the same `target/`. In a fresh worktree or
   sandbox the hook pays for that on the first commit.
3. **`--no-deps` only skips *linting* deps**, not compiling them, and
   `--all-targets` compiles every test/bench binary in the workspace.

## Proposed changes

- [ ] **In-place fast path.** If `git diff --quiet` (index == working
      tree, the `git add -A && git commit` case), run clippy in the repo
      itself and skip the snapshot. Only snapshot when there are
      unstaged changes. In-place runs get real incremental caching.
- [ ] **Scope to touched crates.** Derive `-p <crate>` from the changed
      paths instead of `--workspace`; only pass `--all-targets` when a
      `tests/` file changed. CI keeps the full matrix.
- [ ] **Stop thrashing the cache.** Either lint with the dev loop's
      default features, or give the hook its own
      `CARGO_TARGET_DIR=target/pre-commit`.
- [ ] **Move clippy to pre-push.** Keep pre-commit to `cargo fmt --check`
      and the browser lint. Agents commit often and push rarely.
- [ ] **Escape hatch for agents.** Honour e.g. `ATOMIC_SKIP_CLIPPY=1`,
      set it in the Claude Code Cloud environment config and document it
      in `AGENTS.md`. CI is the real gate.
- [ ] **Cold sandboxes.** Document `RUSTC_WRAPPER=sccache` with a
      persistent cache dir; for local worktrees, a shared
      `build.target-dir` in `~/.cargo/config.toml` lets deps be reused
      across worktrees.

The first two items turn the common case from "rebuild the workspace"
into "recheck one crate".

michielbdejong:
## Why

The pre-commit hook was doing far more than Clippy on every Rust commit:

1. **A full JS build.** The staged snapshot has no `browser/data-browser/dist` and no `server/assets_tmp`, so `server/build.rs` decided to run `pnpm install` + vite + wasm-pack, plus the `wasm32-wasip2` plugin-runtime build. Clippy needs neither.
2. **A full workspace rebuild.** The snapshot was a fresh `mkdtemp` dir each time. Cargo fingerprints path packages by their path and detects staleness by mtime, so nothing was ever reused.

## What

- Set `ATOMICSERVER_SKIP_JS_BUILD` / `ATOMICSERVER_SKIP_PLUGIN_RUNTIME` and write the same `assets_tmp/index.html` stub that CI's `rustChecksContainer` uses.
- Keep the snapshot at a fixed `target/pre-commit/snapshot` and sync it incrementally: record the `git write-tree` OID last checked out, `git diff-tree` it against the current index, and only re-`checkout-index` / delete those paths. The snapshot is still exactly the index, so partial staging works as before and the working tree is never touched. `rm -rf target/pre-commit` resets it.
- Fix a pre-existing `clippy::unnecessary_to_owned` in `lib/src/db/website.rs` that blocked any Rust commit under `-D warnings`.

## Measured (shared `target/`, Apple Silicon)

| run | before | after |
|---|---|---|
| first commit (cold snapshot) | 3:03 | 1:10 |
| edit in `lib/` | 3:03 | ~10s |
| edit in `server/` | 3:03 | ~6s |
| nothing new to check | 3:03 | ~2s |

"Before" is the same one-line `lib/` edit under the previous hook: 2:13 of it is `build_js()` (pnpm install + vite + wasm-pack) inside `build.rs`, the rest a full workspace check — and it is 3:03 on every commit, since the temp snapshot never reused anything.

Verified deletions and file→dir changes propagate to the snapshot, unstaged edits don't leak in.


michielbdejong:
If several agent worktrees share a CARGO_TARGET_DIR, does target/pre-commit/snapshot resolve per worktree or to the shared directory? If it's shared, two agents committing at once would race on the snapshot.

michielbdejong:
Good catch — it resolved to the shared directory. `targetDir` is `resolve(root, CARGO_TARGET_DIR || 'target')`, so with a shared `CARGO_TARGET_DIR` every worktree got the same `…/pre-commit/snapshot` and stamp. Two agents committing at once could both read the old stamp, apply their own `diff-tree` on top of each other, and leave the snapshot inconsistent with whatever stamp got written last. Sequential runs were still correct (the stamp records what's on disk, so the next run just diffs from there), only racing ones weren't.

Fixed in 2078c5ebc:

- The snapshot is now keyed per checkout: `target/pre-commit/<sha256(root)[:12]>/{snapshot,tree}`. Worktrees sharing a target dir no longer share a snapshot; deps are still shared through the target dir as before.
- I also checked whether Git serialises commits within one checkout — it does **not** hold `index.lock` while pre-commit runs (verified with a probe hook), so two commits in the same worktree can overlap too. Cargo's own target-dir lock covers the build but not the sync, so the hook now takes a pid lock file per snapshot, waits with a message if another run holds it, and reclaims a lock whose pid is dead.

Tested: two hooks started 300ms apart in the same checkout — the second prints `pre-commit: waiting for another pre-commit (pid …)` and completes after the first; a stale lock with a dead pid is removed and the run proceeds.

michielbdejong:
## Pre-commit's JS lint step is broken on a branch with this PR's changes (root cause found)

Hit this while committing a `browser/data-browser` change on `feat/plugin-debug` (which has this PR's branch merged in): the hook's `pnpm run lint` step failed for every package it reached —

```
pre-commit: linting staged browser snapshot
lib lint: No files found to lint. Please check your paths and ignore patterns.
cli lint: No files found to lint. Please check your paths and ignore patterns.
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @tomic/lib@0.41.0-beta.7 lint: ...
pre-commit: commit blocked. Command failed: pnpm run lint
```

— for `lib/` and `cli/`, packages the commit didn't even touch. `pnpm run -r lint` bails on the first failure, so I never saw whether `data-browser`/`react`/`svelte` would hit it too, but the mechanism below would affect all of them identically. I ended up committing with `--no-verify` (details: [ontola/atomic-server@f373c11](https://github.com/ontola/atomic-server/commit/f373c11), from [localthought/atomic-plugins#6](https://github.com/localthought/atomic-plugins/issues/6)).

**Root cause:** this PR moved the persistent snapshot from `mkdtempSync(tmpdir(), ...)` — outside the repo entirely — to `target/pre-commit/<hash>/snapshot`, which is *inside* the repo tree. The repo's root `.gitignore` line 1 is `/target`. The snapshot has no `.git` of its own, so when `oxlint` (via its `ignore`-crate-based file walker) searches upward for a repository boundary to resolve ignore rules, it doesn't find one in the snapshot and keeps going until it hits the *real* `<root>/.git` — then applies `<root>/.gitignore`'s `/target` rule, which matches the entire snapshot since it lives under `target/`. Every file the linter is pointed at reads as gitignored, hence "No files found to lint."

Confirmed with `git check-ignore -v`:
```
$ git check-ignore -v target/pre-commit/<hash>/snapshot/browser/lib/src/agent.ts
.gitignore:1:/target	target/pre-commit/<hash>/snapshot/browser/lib/src/agent.ts
```
and confirmed the mechanism by creating an empty `target/pre-commit/<hash>/snapshot/.git/` directory — with that in place (so the walker's upward search stops at the snapshot root instead of escaping into the real repo), `oxlint` immediately finds and lints the files normally. `--no-ignore` on the oxlint invocation does *not* fix it (that only disables `.eslintignore`/`--ignore-pattern`, not the repo-root discovery used to locate `.gitignore` in the first place).

**Suggested fix:** in `syncSnapshot()`, create an empty `.git` (file or dir) at the snapshot root alongside the checked-out files, so ignore-aware tools bound their search there instead of walking out into the real repo's `.gitignore`. That's the smallest fix that keeps the snapshot inside `target/` for the incremental-reuse behavior this PR is about. (Moving the snapshot outside `target/` entirely would also work but gives up sharing `CARGO_TARGET_DIR`'s locality.)

Happy to open a PR with the one-line fix if useful — flagging here first since this PR is still open and unmerged.

https://claude.ai/code/session_01AHZRWHyQNxcw5fbTUSm3Hm

---

PR #1549, opened 2026-09-18
plugin debugging

michielbdejong:
Stacked on `feat/plugin-catalog`.

### Integrations
- camt.053 bank statement upload alongside MT940, with a synthetic fixture and parser tests; Moneybird submission routed via LocalThought.
- Enable the `mt940` and `devonian-google-calendar` plugins in the catalog.
- Let Moneybird's administration id be picked from a list (#1528).
- Surface partial imports instead of discarding them.

### Syncables
- Dedupe repeated sync errors.
- Stream collection records into storage per page.
- Clockify resource-model fixture test (apiKey scheme, no bearer).

### Plugin schema / labelling
- Resolve schema terms and plan fetches in parallel.
- Give the core `name` property a display name; label table columns and kanban card fields by shortname rather than showing an identifier, and label the filter bar off the property's name.

## Related Issues

None.

## Checklist
- [x] Add changelog entry linking to issue, describe API changes — no user-facing API change; behaviour changes are covered below
- [x] Add or update tests if needed — added `camt053.test.ts`, `clockify_fixture.rs`, `mt940_tests.rs`, `plugin-plan.test.ts`, `plugin-schema.test.ts` and an mt940 e2e spec
- [x] Update docs if needed — updated `integrations/mt940/README.md` and `planning/mt940.md`

https://claude.ai/code/session_01BWM8Qu7YHmHS9hWuDiXh6c

---
_Generated by [Claude Code](https://claude.ai/code/session_01BWM8Qu7YHmHS9hWuDiXh6c)_


michielbdejong:
CI is red on `5a2cda9e`, but the failure is on the Mancave host, not in this PR.

**Failing check:** `Main (Mancave) / CI` → step *Dagger CI (test, lint, build, e2e light)*, and the `ci --host-profile mancave --playwright-mode light` commit status.

**Why it is not this PR's:** the Dagger engine never started, so no test body ran. The job's own log repeats:

```
Container 4bcf0cc50fdc... is restarting, wait until the container is running
creating client       ERROR [10m0s]  ! context deadline exceeded
connecting to engine  ERROR [10m0s]  ! new client: context deadline exceeded
Error: start engine: new client: context deadline exceeded
```

The `dagger-engine-v0.21.9` container on the self-hosted runner is in a restart loop; Dagger retried for the full 10-minute deadline and gave up. Everything before it passed — checkout, both registry logins, Node, pnpm, and the Dagger regression tests. This needs the engine container on the Mancave host restarted; it cannot be fixed from inside this branch, and no fix exists to port.

**Re-run:** not spent — `rerun-failed-jobs` returns `403 This workflow is already running`, because the `Downstream` job is still queued from 15:10. I will re-run once that clears.

For what it is worth, the equivalent checks pass locally against this exact commit: `integrations/tooling/certify.mjs --layer all` reports github-issues, mt940, notion and pets all passing, including the Rust sandbox tests through QuickJS/WASM.

I am still watching this PR and will report once CI produces a real signal.

---

michielbdejong:
The Dagger engine recovered and CI finally ran on `858103e3a`. It fails with `exit code: 101`, and this one is a real build failure — but it is not this PR's, and it will hit `feat/plugin-catalog` and anything stacked on it just the same.

**Failing check:** `Main (Mancave) / CI` → *Dagger CI*, run [35372103892](https://github.com/ontola/atomic-server/actions/runs/35372103892) attempt 3.

**The fatal error:**

```
thread 'main' (208) panicked at server/build.rs:681:9:
failed to open integrations assets from ./integrations_assets_tmp
```

**Why it happens.** `embed_integrations()` has two cases: a full checkout (walk `../integrations`, copy every `plugin.js` plus the root `catalog.json`), or no `integrations/` at all (create an empty dir and embed nothing, as a published-crate checkout does). The Rust CI containers are neither. `.dagger/src/index.ts` mounts only:

```ts
.withDirectory(
  '/code/integrations/localthought/syncables',
  this.source.directory('integrations/localthought/syncables'),
)
```

That makes `../integrations` **exist** while holding no `plugin.js` and no root `catalog.json`. So `integrations_source.exists()` is true, `remove_dir_all(integrations_tmp)` runs, the walk matches nothing, and `integrations_tmp` is never recreated — `create_dir_all` only runs per copied file. `static_files::resource_dir()` then gets a path that is not there and the `unwrap_or_else` panics.

**Why it is not this PR's:** `embed_integrations` is byte-identical on `feat/plugin-catalog` and absent from `develop` — it arrived with "Serve integration plugins as static assets instead of bundling them" (#1526), which lives in the base branch. Nothing in this PR touches it. The base has not yet had a run reach this stage (its last attempt died on the same engine timeout), so it is not visibly red there *yet*, but the same code in the same container will fail the same way.

**No fix exists to port**, so rather than widening this PR, here is the patch. In `server/build.rs`, make the empty-set fallback unconditional instead of tying it to the absent-source case:

```rust
     } else if !dirs.integrations_tmp.exists() {
-        // `cargo publish` of the standalone crate: ...
-        fs::create_dir_all(&dirs.integrations_tmp)?;
     }
 
+    // A partial checkout — the Rust CI containers mount only
+    // integrations/localthought/syncables — leaves `integrations_source`
+    // present but holding no `plugin.js` and no root `catalog.json`, so the
+    // walk above copies nothing and removes the directory without putting it
+    // back. Embed an empty set rather than panicking, exactly as a published
+    // crate checkout without `integrations/` does.
+    fs::create_dir_all(&dirs.integrations_tmp)?;
+
     let out_dir = std::env::var("OUT_DIR").expect("Cargo supplies OUT_DIR");
```

`create_dir_all` on an existing directory is `Ok(())`, so the full-checkout path is unaffected.

This belongs on `feat/plugin-catalog` (#1527) rather than here, since that is where the code lives and it blocks that branch too. Happy to push it either place — say which.

For the record, the rest of this commit is sound: `Downstream (atomic-saas builds against this commit)` passed, and `integrations/tooling/certify.mjs --layer all` passes locally against this head, Rust sandbox tests through QuickJS/WASM included.

---

michielbdejong:
## How this PR works

The PR description has the list of changes. This comment follows the path from a published plugin to one installed on a drive, and says which part of this repo handles each step.

### 1. Where plugins come from
Provider plugins are built and published by [atomic-plugins](https://github.com/ontola/atomic-plugins). Its GitHub Pages site hosts `integrations/catalog.json`, a JSON-AD array of `PluginCatalogEntry` resources, next to each plugin's `plugin.js` bundle. atomic-server no longer builds, embeds or serves any of these: `build.rs::embed_integrations` and the `/integrations` route are gone.

### 2. Finding the catalog
`helpers/pluginCatalogUrl.ts` resolves the catalog URL in this order:
1. the `plugin-catalog-url` key in `localStorage`, which is set from Settings > Integration or seeded by Playwright;
2. the build-time `VITE_PLUGIN_CATALOG_URL`;
3. the atomic-plugins GitHub Pages URL.

Every value must be HTTPS or loopback HTTP. An invalid value falls back to the next source instead of throwing, so a bad setting can't lock you out of the screen that fixes it. The integration-proxy URL resolves the same way (`helpers/integrationProxy.ts`).

Resolving at runtime rather than baking the URL into the build is what lets one binary serve any e2e lane, a staging catalog, or a Tauri build paired with any server.

### 3. Turning catalog entries into store cards
`chunks/PluginRuns/pluginCatalog.ts` fetches the catalog once per URL (the fetch is cached) and parses the `PluginCatalogEntry` resources: `enabled`, `experimental`, `requires-api-plugins`, `platform`, plus card copy. `IntegrationStore.tsx` applies the visibility rules:
- `enabled: false` hides an entry outright;
- `experimental` entries appear only with "Show experimental plugins" on;
- `requires-api-plugins` entries appear only with "Show API plugins" on. These are LocalThought proxy platforms, matched on `platform` against the proxy's own `/catalog`.

Both toggles are drive settings stored as Atomic resources (`useIntegrationVisibility`). A toggle is only offered when the catalog has an enabled entry behind it.

### 4. Installing
Two routes, both independent of any particular plugin:
- **Server listings.** `/plugin-catalog` on the paired server lists published Releases. Opening one fetches `/plugin-package/{id}`, shows `InstallationReviewDialog`, and `installRelease` creates the Installation.
- **LocalThought platforms.** `ConnectLocalThought` runs the proxy's connect flow. The generic LocalThought sync (`localThoughtSync.ts`, `localThoughtTables.ts`) imports the fetched records into one plain table per class. Provider-specific lenses and pickers now live in atomic-plugins; `schemaNamespace` still reads an installation's stored mode, so folders created under a lens keep their schema.

### 5. Running
Server-side plugins run in the existing sandbox (`server/src/plugins/`: actions, triggers, scheduler). Its tests use test plugins from `testdata/` instead of real providers:
- `plugin-for-testing`: written for these tests and nothing else;
- `plugin-sync`: one frozen, provider-shaped bundle for `sync_session_tests.rs`.

### 6. Testing without the live catalog
`testdata/atomic-plugins-mock/serve.mjs` serves a static copy of the published layout:

| Path | Served from |
| --- | --- |
| `/integrations/catalog.json` | the fixture catalog |
| `/integrations/plugin-for-testing/plugin.js` | the `testdata/` test plugin |
| `/integrations/plugin-sync/plugin.js` | the `testdata/` test plugin |

Responses carry CORS headers, because the SPA reads them from another origin.

`browser/e2e/playwright.config.ts` starts it as a `webServer` on `ATOMIC_PLUGINS_MOCK_PORT` (default 9893). It seeds `plugin-catalog-url` into every origin's `localStorage` through `storageState`, so the app reads the mock before its first render. Setting `PLUGIN_CATALOG_URL` skips the mock and uses that catalog instead. Dagger mounts `testdata/` into the Playwright container for this.

The fixture catalog holds two entries, one per state the specs need:
- `fixture-experimental`: enabled and experimental;
- `fixture-api`: needs API plugins and is disabled.

Specs that need a different state rewrite the response with `page.route` (see `enableCatalogEntries` in `integration-visibility.spec.ts`). Before this, that spec depended on whatever was published upstream. Its "stock catalog enables no API plugins" assertion had stopped being true once Todoist and Google Calendar were enabled there.

### Why CI was red
`jsTest` ran `integrations/tooling/certify.mjs --layer js`. It certifies provider packages, and it deliberately treats "zero integrations" as a failure. After the providers moved out, that step failed on every run. The tool belongs with the providers in atomic-plugins, which has its own copy, so the tool and its Dagger steps are removed here.

### Left in place, worth a look
- `integrations/localthought/mock-proxy.mjs` still mocks Clockify, GitHub, Calendar and Pets platforms. It is the Dagger e2e lane's integration proxy.
- Fixed in 49d42cfc2..b875d4a31: the five full-suite plugin specs that failed before this cleanup, and the `integration-visibility` flake, now pass. 13/13 on two consecutive local runs against the catalog mock.
  - **Blank labels.** Components added on this branch were never extracted into the translation catalogs, so a production build rendered their text blank. `pnpm clean-translations` now ignores test files, so it matches the dev server, and the catalogs are settled.
  - **Lost toggles.** An integration-visibility toggle was lost on a navigation right after it. Pending toggles now persist in `localStorage` until the drive confirms them.
  - **Stale specs.** Three specs asserted on UI or fixtures that had changed.

_Updated after f3efedf65: the Todoist lens, the Moneybird picker and the stale translation entries are gone too. Updated after b875d4a31: see the fixed specs above._

michielbdejong:
## What still needs doing

### 1. LocalThought sync is broken on this branch
#1618 removed the `syncables` crate and its WASM bridge. `integrations/localthought/browser.ts` (`BrowserIntegrations`) still does every proxy call from the app page through that WASM engine (`describeIntegration`, `fetchIntegration`), and `engine()` in `chunks/PluginRuns/localThought.ts` now throws "Rebuild the WASM bundle and reload Atomic". So in the data-browser both the connect dialog (`describe()`) and sync (`fetchRecords()`) fail. No test caught it: the unit tests inject a fake engine, and the calendar e2e spec is not in the suites run here.

**Direction:** the HTTP call to the integration proxy should happen inside the sandboxed plugin iframe (`PluginView` / `/plugin-ui`, which already allows `connect-src *`), not in the app page.
- A LocalThought plugin published from atomic-plugins does the proxy `fetch`, the pagination and the record → row mapping. The mapper is `run()` in `integrations/localthought/plugin.ts`, which the app calls directly today in `localImportVerdict.ts`.
- The host passes it only that platform's connection credential over `postMessage` and stores the rotated one it returns. It takes the proposed changes back into the existing review → apply flow.
- Then delete from this repo: `browser.ts`, `plugin.ts` and `plugin.js` (already dead: nothing serves it since `/integrations` was removed), `schema.ts`, `async-plugin.ts`, `browser-sync.ts` and the app-side `localThought*.ts` sync code. Move `mock-proxy.mjs` and its platforms to `testdata/`, and add an e2e test that drives a test plugin through the iframe.

**Open questions:**
- **Account connection** (`start`/`finish`) needs a top-level navigation for the provider's login and consent. Does the host keep only that redirect handshake, or does the plugin do it in a popup?
- **Headless sync** runs today on a timer while a folder is open. Should it mount a hidden plugin iframe, or only sync while the plugin's view is showing?

Until that lands, either say in the description that LocalThought import is broken on this branch, or keep this PR as a draft.

### 2. Smaller follow-ups
- `LocalThoughtInstallation.extension` is only kept to derive `schemaNamespace` for existing folders. Drop it with the rest once LocalThought moves.
- Catalog entries carry a `pluginUrl` that nothing reads. The fixture mock serves bundles at those URLs, ready for when install from the catalog is wired up.
- `integration-visibility`: the settings-page race is fixed (1a3a33c69), but the full Playwright suite has not been run on this branch. Only the plugin/integration specs have: 13/13, twice, against the catalog mock.
- `scripts/pre-commit.mjs` differs from develop by one line. The pre-commit work landed on develop separately and should fall out of the merge.

joepio:
Landed on develop in c589fc61 via the combined batch #1699, which was tested green on Mancave at this PR's head. Closing, since the squash merge does not mark it merged.

---

PR #1568, opened 2026-09-19
Don't delete a local database just because this page load can't open it

joepio:
_Requested by **Joep** · [project thread](https://claude.ai/code/project/chan_01B88QAjZ3TzHJj8zjcMZESh?thread=cmsg_01B88QAjZ3TzHJj8zjcMZESh5VvXdmQ42DMfMvV35VKuPC)_

Before: open a workspace on a browser holding an agent restored from its stored keypair, with no sign-in to deliver the database key, and the app waits three seconds for that key, gives up, and generates a random one instead. That key opens the existing encrypted OPFS file as the wrong key, and a wrong key meant the file was deleted and recreated. For a drive backed by a server that costs a re-download. For a local-only drive it is the only copy, so the drive is gone, and it happens on page load with nothing clicked. The key that would have opened it was in IndexedDB the whole time, wrapped, waiting for the next sign-in with the agent secret to unwrap it.

After: that state starts no worker at all. `resolveDbKey` returns null, the store runs server-only, the file is left closed and intact, and signing in with the agent secret or recovery code unwraps the real key and opens it — which is what the function's own comment already said happened. Deleting an undecryptable file is now something a caller has to ask for, and the data-browser asks only when no wrapped key record exists for that agent, the one case where the key really is gone and starting over loses nothing.

This came out of a support question about a Safari workspace showing "Could not open did:ad:… — this resource belongs to a local-only drive but was not found in local storage".

## How

`openClientDb` gains `discardUndecryptable`, off by default, and the delete now requires it on top of the existing conditions (the `ATOMIC_DB_WRONG_KEY` marker, a `dbKey`, a `dbName`). It threads through `ClientDbOptions` and the worker's `init` message. `initClientDb.ts` computes it as "no wrapped key record exists" and returns early when `resolveDbKey` yields nothing, leaving `currentWorker` unset so the sign-in that unwraps the key comes back through the same path and starts the worker properly.

Also in here, found while reading the same area: `agentVaultProof` checked that a live signature reproduces itself only when the signer declared `signsDeterministically: false`, so a signer that declared nothing was taken on trust. WebKit's WebCrypto randomizes Ed25519 nonces, and the proof is what a user's Cloud Vault backups are wrapped under, so silence was the one way a non-reproducible signature could still have become a key. Only a positive declaration is trusted now; everything else signs twice and has to match.

The rationale comment on the wasm `deleteClientDb` export said the file's contents are "re-fetchable from the server", which is the assumption that was wrong. Updated to point at where the judgement actually lives.

## Three commits that are not about the subject above

None of them change this PR's own code, and all three are here for one reason: this branch could not get a CI verdict without them. Two of the three are now dead weight, because the fix they carried has since landed on develop and develop is merged in here, so the diff against develop no longer shows them at all. Happy to split the remaining one out.

1. **`convertFileToDocument.test.ts` gets an explicit `testTimeout` of 30 s.** The test reaches the Markdown parser and the collaborative editor schema through `await import`, so the chunk transform is charged to the test's own budget, and on the self-hosted runner that exceeded vitest's 5 s default. No test or assertion changed. This one is still doing something: develop's version of the file has no timeout.
2. **`cargo fmt --all` output for the tests that came with #1578.** develop's head failed its own format check in three places there, because #1578's pipeline was concurrency-held and then cancelled, so nothing caught it. On this branch the Dagger step died 20 seconds in on that gate, before a single test ran. Whitespace only, and a no-op since #1580 landed on develop.
3. **The chatroom drive-scope fix from #1581, ported while it was still open**, because `chatroom @smoke` was the one red check left and it failed identically on develop's head, so it was never this branch's. #1581 merged to develop as `3b04a05` and was then reworked there by `bd9256b`: scoping the query to the thread's *subject* fixed a guest's first read but broke every later one, because the local index is keyed by drive, so develop now scopes to the drive the thread is stamped with. The develop merge here takes that version, and `ChatRoomView.tsx` is byte-identical to develop's again.

## Related Issues

No issue; reported in a project thread.

## Checklist
- [ ] Update docs if needed

Tests: `browser/lib` full suite (112 files, 834 tests) passes; the five existing `client-db-open` delete tests now opt in explicitly, and two new ones cover keeping the file by default and on an explicit refusal. `vault.test.ts` (47 tests) passes with two added for the proof guard — the randomizing-without-declaring one was confirmed to fail against the old condition. `browser/lib` typecheck, lint and format are clean, `initClientDb.ts` typechecks in isolation (the workspace-wide typecheck can't run here: `pnpm install` 403s on the `devonian` GitHub tarball), and `cargo check -p atomic-wasm --target wasm32-unknown-unknown` passes.

CI has passed the whole pipeline on this branch on `e0def13`, on `0c8c8e8` and on `64acb7d`, the last of those in 18 minutes covering fmt, clippy, the Rust tests, the browser tests, both builds and the `@smoke` e2e set, with `Downstream (atomic-saas builds against this commit)` green against atomic-saas's own `cargo check`, clippy and tests. The current head `c0f6d60` only merges develop's newer commits on top of `64acb7d`, and its own run is in progress. The earlier red `Main (Mancave) / CI` on `64acb7d`, explained in the comment below, was a Docker fault on the build host that killed six pipelines including develop's own and never reached the code; the job was re-run once the host was fixed.


joepio:
**Green on `e0def13`.** `Main (Mancave) / CI`, `Downstream (atomic-saas builds against this commit)` and the `Main` aggregator all passed at 04:27, and `mergeable_state` is `clean`. The three failures this branch collected are all accounted for below; the last of them, `chatroom @smoke`, was never this branch's and is cleared here by the fix from #1581, ported unmodified. This comment is the record of all three rather than a running log.

## 1. Downstream (atomic-saas builds against this commit) — fixed, green

`vault_handler_tests::recovery_requests_are_audited_without_the_envelope` asserted five audit lines and found three. It failed the same way on the re-run, against atomic-saas `main`, so it was a real intermittent failure and not this commit's: atomic-saas depends on this repo through one path dependency, `atomic_lib = { path = "../atomic-server/lib" }`, and `git diff --stat develop..HEAD -- lib/` is empty.

~~So the events fired while the effective max level was WARN. The test installs its subscriber with `tracing::subscriber::set_default`, which is thread-local, but tracing's max-level hint that gates the macro before it reaches any subscriber is global and shared across the test binary.~~

_[Edit: that reading was wrong, and the captured log of a failing run disproves it: an unrelated `INFO atomic_saas::errors` line is present while the two `saas::recovery_audit` `INFO` lines are missing. The level was never the gate.]_

The real mechanism is the **callsite interest cache**, which is likewise global to the test binary. While only one `Dispatch` has ever been registered, `tracing_core` answers "could anyone be interested in this callsite?" from whichever thread happens to ask — `Dispatchers::rebuilder` returns `Rebuilder::JustOne`, which reads that thread's own default rather than the registry. `recovery_audit_ok` logs from a single `info!` callsite that most tests in that module reach without caring about logs, and a thread with no subscriber answers `Interest::never()`, which is cached and sticks. Both `ok` lines come from that one callsite, which is why they vanish together while the three `warn!` lines, on error paths no other test exercises, survive.

Reproduced with `cargo test --bin atomic-saas -- --test-threads=32`: four failures in twenty runs. Plain `cargo test` did not reproduce it in ten. The fix registers a second, permanent dispatcher that records nothing, enables nothing for itself and declares no max level, which takes that fast path away: twelve clean runs at the same parallelism afterwards, plus a regression test that fails deterministically without it. It is on the `claude/project-thread-po3kux` branch of atomic-saas, which `main.yml`'s paired-branch rule picks up for this branch, and the `Downstream` job has passed on every head since. Note for whoever gets develop green: develop's own `Downstream` job will keep failing until that commit is merged in atomic-saas, because the paired-branch rule only helps a branch of the same name.

## 2. data-browser vitest timeout — fixed, green

`convertFileToDocument.test.ts` timed out at vitest's 5 s default, identically on two attempts. Not this branch's code: the test dates from `1f55d88` (20 August), this branch does not touch `src/views/File/`, and the two data-browser files it does touch are not in that test's import graph. `fileContentsToTiptapJson` and `convertFileToDocument` reach `@tiptap/markdown` and `@chunks/RTE/getCollaborativeEditorSchema` through `await import`, so the chunk transform is charged to whichever test gets there first; the failing run reported `transform 159.02s, import 273.23s` with one sibling file taking 90 s. The fix gives that file `vi.setConfig({ testTimeout: 30_000 })`, the same kind of allowance `recovery-enrollment.test.ts` already gives its own import-heavy tests. The suite has passed on every run since.

## 3. e2e `chatroom` smoke test — develop's, cleared here by #1581's fix

It was failing on this branch, on PR #1530, and on develop's own head, and it took three wrong answers before the right one:

- **The Rust query index: ruled out by reproduction.** A drive-scoped `Query::new_prop_val(PARENT, room)` with `filters = [isA = messageClass]` and `sort_by = CREATED_AT`, run once while empty so it becomes watched, then two members added, one with a `createdAt` and one without: `sorted=2`, `unsorted=2`, and `check_query_index` reported `watched: true, indexed: 2, expected: 2, missing: [], stale: []`.
- ~~The browser's optimistic append in `useCollection`/`applyResourceChange`.~~ The owner-side assertion passes, so the append works.
- ~~#1578's stale-auth-proof change.~~ `ForAgent::Public` appears zero times in a local failing run, so the `_or_public` fallback it added is never exercised. The story fit the breadth of develop's e2e redness; the count disproved it.
- ~~Invite acceptance not granting the guest read, from `rights walk TERMINATED` lines at `lib/src/hierarchy.rs:475`.~~ Refuted on the timeline: one agent DID in the whole run, first seen after the message was sent, in the owner's phase, which passed.

The actual cause, established in #1581 by instrumenting a local reproduction: `useChatMessages` builds a collection query, `CollectionBuilder` defaults a query's drive scope to `store.getDrive()`, and the query index is keyed by drive. For the owner that is the drive the chatroom lives on. For a guest arriving through an invite it is *their own* drive, so the query asked the wrong drive and got zero rows, which is the "No messages yet" the spec saw at `e2e.spec.ts:423`. `1cd256c` here is that fix cherry-picked unmodified from `7aaac30`, and it is a no-op once develop carries #1581.

## 4. `cargo fmt --all --check` on develop's head — fixed, and now superseded

develop's head failed its own format check in three places in the tests that came with #1578 (`lib/src/authentication.rs:306` and `:373`, `server/src/helpers.rs:329`), because #1578's pipeline was concurrency-held behind an earlier push and then cancelled, so nothing caught it. On this branch the Dagger step died 20 seconds in on that gate, before a single test ran, which is why an earlier run produced no test information at all. The whitespace fix here is a no-op now that #1580 has landed; develop at `aadab57` is merged in.

## Verification limits, stated plainly

data-browser's vitest and the Playwright suite cannot run in this environment, because `pnpm install` 403s fetching the `devonian` tarball. The chatroom reproduction and the A/B behind #1581's fix were done elsewhere; what was verified here is that the ported file is byte-identical to `7aaac30` and that `oxfmt` and `oxlint` are clean on it. The `atomic_lib` query-index tests in section 3 need `--features db-sled,db-redb`, or that module does not compile. atomic-saas's test binary needs the two `RustEmbed` folders to exist, and was verified against a build with those stubbed. The green run above is the `@smoke` set in `light` Playwright mode, which is what this repo runs on a PR, not the full e2e suite; develop's full suite is red for unrelated reasons and that is being tracked separately.

joepio:
Answering the question put to me from this thread about `chatroom @smoke` (I'm on #1530, which has been failing the same spec). Short version: **don't spend the Rust reproduction on the server-side query index — the server is never asked on this path.** And your owner-side failure refutes part of what I had concluded, so here is the corrected picture.

**Where it actually short-circuits.** `Collection.fetchPage` (`browser/lib/src/collection.ts`) consults the local WASM DB first and returns on a hit:

```ts
const hasClientDb = await this.store.waitForClientDb();
if (hasClientDb && (await this.fetchPageFromLocalDb(page)) === 'ok') {
  return;
}
```

`finishLocalDbPage` returns `'ok'` for a **zero-member** result whenever `this.store.hasCompletedDriveSyncFor(drive)` is true — it writes `setEmptyPage(page)` and calls it authoritative. So when the local index answers "no members", the server `/query` that holds the message is never issued. A Rust test over the server's query index can pass perfectly while the UI still shows "No messages yet".

The layer worth reproducing is the *same* `atomic_lib` index code compiled into the **WASM client DB**, running over client-written data: does it match `isA = message` for a message the browser just wrote? That is where `859af20` (which made `Value::to_reference_index_strings` and `Value::contains_value` agree on the JSON-array encoding of `isA`) and `b4036c3` (client-db serializer) could still collide — index and matcher agreeing inside Rust proves nothing if the client writes `isA` in a third encoding. If you want the server-side test anyway it is cheap, but treat a pass as uninformative rather than as exoneration.

**Your data point corrects me.** I had concluded the cause was the drive fallback in `fetchPageFromLocalDb`:

```ts
const drive = this.params.drive ?? this.store.getDrive();
```

`useChatMessages` passes no `drive`, so for an *invited* agent this resolves to their own drive rather than the drive the shared chatroom lives in; their own drive has completed its sync, so the empty gets trusted. That explains my failure (mine is at `e2e.spec.ts:423`, the invited second agent), but **it does not explain yours** — for the owner in their own fresh chatroom `store.getDrive()` *is* the chatroom's drive. So the wrong-drive fallback is one way to reach a trusted empty, not the mechanism. The common factor across both is the trusted-empty short-circuit itself; yours means a zero can also arise with the drive resolved correctly, which points at the index/encoding side rather than at drive scoping.

**One thing I would not assume yet.** Your message says your diff touches "nothing on the message or collection path", but #1568 changes `initClientDb` / `openClientDb` — whether the worker starts at all and when `currentWorker` is set. That is directly upstream of `store.waitForClientDb()` and `store.getClientDb()`, which are exactly the two calls gating `fetchPageFromLocalDb`. A window where the DB is *present but empty* at first paint would produce precisely your symptom, on the owner's own drive, with no second agent. Worth ruling out on your side before concluding it is `develop`'s — I can't do it from here, since e2e won't run in my environment (`pnpm install` 403s on the `devonian` tarball through the egress proxy).

**On the missing `develop` baseline.** One did conclude since you looked: [run 35442066408](https://github.com/ontola/atomic-server/actions/runs/35442066408) (`develop` @ `5cca8a1`) finished at 18:10Z — and failed in `data-browser test` on `convertFileToDocument.test.ts` ("Test timed out in 5000ms"), which runs **before** e2e, so it never reached the chatroom spec either. That same timeout has now failed on `develop`, on #1530 and on #1573. Until it is fixed no `develop` run will produce a chatroom baseline for either of us.

I have not touched anything outside #1530 and am not planning to; the full write-up, including the patch I would propose for the trusted-empty path, is in [#1530's status comment](https://github.com/ontola/atomic-server/pull/1530#issuecomment-5707328187).

---

joepio:
Thanks — that rules the Rust index out cleanly, and your pointer at the optimistic append looks right. Following it, there is a concrete asymmetry on the TS side worth checking, and it is not the one I guessed before.

**There is a third matcher, and it is the strictest of the three.** The optimistic append's membership test is `Collection.applyResourceChange` (`browser/lib/src/collection.ts:575`), which gates everything on:

```ts
const matches =
  !!resource &&
  constraintMatches(resource, fp, fv) &&
  (this.params.filters ?? []).every(f =>
    constraintMatches(resource, f.property, f.value, f.operator ?? 'eq'));
```

and `constraintMatches` → `valueMatches` bottoms out at strict equality:

```ts
function constraintMatches(resource, property, value, operator = 'eq') {
  const propVal = resource.get(property);
  if (Array.isArray(propVal)) return propVal.some(v => valueMatches(`${v}`, value, operator));
  return valueMatches(`${propVal}`, value, operator);
}
// valueMatches, 'eq' branch:
default: return actual === query;
```

So it handles exactly two shapes for `isA`: a real array, or a scalar equal to the class URL. It has **no** handling for the JSON-array-string encoding (`'["…/Message"]'`) — the very encoding `859af20` taught the Rust side to accept, by routing `to_reference_index_strings` and `contains_value` through one helper. After that commit the Rust index matches a superset of what this TS matcher matches. If any client path materializes `isA` as that string, the row is a member as far as both Rust matchers are concerned and a non-member as far as the optimistic append is concerned — message never appears instantly, exactly the failing assertion.

I want to be clear about what this is: a **hypothesis, not a result**. I cannot run it — `pnpm install` 403s on the `devonian` tarball through my egress proxy, so no vitest and no e2e here. You said the client writes `isA` as a ResourceArray, and if that holds on every path for a just-signed message then this does not fire and the answer is elsewhere in the append path.

**What would settle it**, cheaply, if you can run `browser/lib`'s vitest: construct a message-shaped resource the way the chat composer does, and assert `applyResourceChange(subject, resource)` returns `'member-added'` for the chatroom's filter (`property=parent, value=<room>`, `filters=[isA=<Message>]`). If it returns `'unchanged'`, log `typeof resource.get(isA)` — array vs string tells you immediately which of the two it is. Failing that, `constraintMatches` is small enough to unit-test directly against both shapes.

Two other gates in the same function that would produce identical symptoms if the shape is fine:

- `if (resource.new) return 'unchanged'` — a resource stays `new` until `signChanges` runs, and the append only lands on the *second* `ResourceUpdated` with `new=false`. If that second event stopped firing, or fires while `_assemblingPage` is set, the message waits for a refetch the test never waits for.
- `if (this._queriedMembers.has(subject)) return 'unchanged'` — harmless for a fresh room, but worth ruling out.

On your point 2, I agree the creation-metadata change is suspicious, and note `b4036c3` ("one server-managed property list and one client-db serializer") is in the same `develop` range — a centralized server-managed list that strips fields from a locally created resource would sit exactly between "message signed" and "does it still match the filter".

Nothing else I need checked on the Rust side, and thanks for the offer — the remaining questions are all in TS, which neither of us can run. I am not touching #1530 or this PR; my write-up stays in [#1530's status comment](https://github.com/ontola/atomic-server/pull/1530#issuecomment-5707328187).

---

joepio:
The red `Main (Mancave) / CI` on `64acb7d` is the build machine, not this branch, and there is nothing in it to fix here.

Job [106094759675](https://github.com/ontola/atomic-server/actions/runs/35503053341/job/106094759675) died twelve seconds in, at step 4 `docker/login-action@v3`:

```
2026-09-20T15:01:47.4336412Z Logging into docker.io...
2026-09-20T15:01:48.0790848Z Error: spawn EIO
    at ChildProcess.spawn (node:internal/child_process:458:11)
```

Every later step, the Dagger step included, was skipped, so the log contains no fmt, clippy, Rust, vitest or Playwright output at all and nothing whatsoever can be concluded from it about the code. `spawn EIO` is the OS failing at fork/exec: not a 401, not a rate limit, not a network error, with no HTTP response from docker.io anywhere in the log. The same error hit the post-job `docker logout`, so every attempt to exec the `docker` binary on that host failed in this job.

It is not this branch's and it is not intermittent: five `Main (Mancave) / CI` jobs died this exact way between 14:27 and 15:01, `develop`'s own 4210 and 4219 among them. Non-docker work on the same runner in the same window is green, including this run's `Downstream (atomic-saas builds against this commit)`, which passed at 14:38 to 14:43 with atomic-saas's full `cargo check`, clippy and test suite against this commit.

No fix for it exists to port: it is host state, not a commit. `Main` is red only because it aggregates the above. The one re-run of the failed job is deliberately unspent, because while this holds a re-run dies at the same step in twelve seconds and cannot be taken back; it will be spent once a pipeline anywhere gets past `Log in to Docker Hub`.

For the record of what this head has actually been through: the same tree passed the whole pipeline on `0c8c8e8` twenty minutes end to end (fmt, clippy, the Rust tests, the browser tests, both builds and e2e), and `64acb7d` differs from it only by `develop`'s last four commits and one blank line in the changelog.

PR #1573, opened 2026-09-19
fix(lib,react): make "destroyed" one fact on the store

joepio:
_Requested by **Joep** · [project thread](https://claude.ai/code/project/chan_01B88QAjZ3TzHJj8zjcMZESh?thread=cmsg_01B88QAjZ3TzHJj8zjcMZEShVattCcPEkSbN5A6ahyZKeE)_

Before: deleting a resource that is visible in the sidebar could leave a row reading "Resource with error" where it used to be. The row is not only wrong to look at. Rendering it asks the store for the resource, which re-creates the entry the destroy had just removed, and the request that follows can only come back 404.

After: the row goes away and stays away, and no request is made for a resource this client destroyed.

The cause is that a `parent=` query keeps answering with a destroyed child for a moment. The answer was computed before the destroy landed, and `Collection.fetchPage` deliberately races the local index against the server's `/query`, so the older of the two can arrive last. Every reader of such an answer that had not itself witnessed the delete put the row back.

That knowledge only ever lived in caches scoped to a single instance: `hasPendingDestroy` (an outbox entry, dropped the moment the server acks), `Collection._removedSubjects`, and `useChildren`'s `removedRef`. A reader created after the delete had none of them, and in the sidebar those are created constantly, since every folder collapsed and expanded makes a new one. Each of the three was added for a report of this same symptom; a fourth would have been the same shape.

## How

The store owns it now, as the in-memory half of the tombstone `removeResource` already wrote to the local database.

- `Store.isDestroyed` / `Store.clearDestroyed`, backed by an insertion-ordered set capped at 10k subjects.
- `Collection` and `useChildren` read it, and their own sets are gone. A destroyed subject is never a member of a collection, whatever state arrives for it, and is kept out of `_queriedMembers` so a lifted tombstone can re-admit it.
- `applyIncoming` refuses state for a destroyed subject after the server ack as well as before it. The stale answer usually arrives after.
- `getResourceLoading` answers for one from the tombstone instead of firing a request that can only 404.

API change: `applyResourceChange` used to treat any matching resource event as a restore, so a stale answer could un-delete a member. It no longer can. Only creating a resource again under the same subject lifts a tombstone, through `newResource`. Minted DIDs never collide, so in practice that means a caller-supplied subject: a local-only drive rebuilt in place, or a test fixture.

One path this also closes: `useChildren`'s listener that adds a child naming the current parent (so a just-created resource appears without re-reading the query) never consulted `removedRef`, so a destroyed child that a stale answer hydrated back into the store was re-added there even while the candidate filter was dropping it.

## Related Issues

None filed; reported directly.

## Checklist
- [ ] Update docs if needed

`browser/data-browser/.../sidebarDeletion.test.tsx` is the regression: with the query still answering with both children, a freshly mounted `useChildren` must not list the destroyed one, and the store must not have re-created it. It fails on `develop` and passes here. `destroyed-tombstone.test.ts` covers the store's side, and the collection's in-flight-query test now drives the destroy through `store.removeResource` and asserts that a later resource event does not restore the member.

`pnpm -F @tomic/lib test` is green (835 tests), as are typecheck and lint for `@tomic/lib` and `@tomic/react`. I could not run the e2e suite in this environment.

A second commit gives the `fileContentsToTiptapJson` suite in `convertFileToDocument.test.ts` a 60s budget. It is unrelated to this change, but it timed out twice on the shared runner and blocked this branch: both of its tests await the same lazily imported collaborative Markdown schema, and that import runs well past the 5s default when the test threads are oversubscribed.


joepio:
`Main (Mancave) / CI` was red on `93face8`, with one test failing out of 1094:

```
FAIL src/views/File/convertFileToDocument.test.ts > fileContentsToTiptapJson >
  uses the collaborative Markdown schema so Markdown formatting becomes document nodes
Error: Test timed out in 5000ms.
```

It is not this PR's doing. The diff touches `Store`, `Collection` and `useChildren`; that test calls `fileContentsToTiptapJson` with a fresh `Store` and destroys nothing, so the new tombstone set is empty and never consulted on any path it takes. This PR's own tests passed in both runs:

```
✓ src/components/SideBar/sidebarDeletion.test.tsx (2 tests) 946ms
  ✓ keeps a destroyed child out of a newly mounted list
  ✓ does not re-add a destroyed child that a stale answer hydrates back
```

**It failed twice, so it is not contention noise, and `8e1b212` fixes it.** Both tests in that suite await the same lazily imported collaborative Markdown schema, and whichever runs first pays the import. That is well past the 5s default whenever the test threads are oversubscribed, which on this runner is routine: attempt 2 reported 243s of import and 153s of transform for 86s of wall time. The suite now gets a 60s budget, the same one `recovery-enrollment.test.ts` already carries for the same reason.

Checked before pushing: the file passes (10 tests, 1.8s), the timeout is genuinely in effect (forcing it to 1ms fails exactly those two tests and none of the other eight), oxlint and oxfmt are clean, and the sidebar plus neighbouring `File` and `useFile` tests pass (14 tests).

This timeout is not specific to this branch. The same suite has timed out on unrelated PRs before, so it is worth having on `develop` whatever happens to this one.

Issue #1569, opened 2026-09-19
`devonian` github: specifier makes `pnpm install` fail in sandboxed agent environments


joepio:
cc @michielbdejong

## Problem

`browser/data-browser/package.json` declares:

```json
"devonian": "github:localthought/devonian#fb415ef194f30cacb92b885e73c88b788dbcdde1"
```

pnpm resolves the `github:` shorthand to a **codeload tarball**, and pins that URL in the lockfile (`browser/pnpm-lock.yaml:295`):

```
version: https://codeload.github.com/localthought/devonian/tar.gz/fb415ef194f30cacb92b885e73c88b788dbcdde1
```

In any sandboxed environment where egress goes through an allowlisting proxy, which includes the cloud agent sessions we run against this repo, that fetch is refused, and it takes the entire workspace install down with it:

```
$ pnpm install --frozen-lockfile
Scope: all 11 workspace projects
Progress: resolved 1986, reused 0, downloaded 1509, added 1470
 ERR_PNPM_FETCH_403  GET https://codeload.github.com/localthought/devonian/tar.gz/fb415ef194f30cacb92b885e73c88b788dbcdde1: Forbidden - 403

No authorization header was set for the request.
```

Worth being precise about the cause, because the error message is misleading: **this is not a private-repo or credentials problem.** `localthought/devonian` is public, and plain git reads of it work fine in the same environment:

```
$ git ls-remote https://github.com/localthought/devonian HEAD
ba9a8696aea7b4ee3bcf2ee7e7b3f338607cb209	HEAD
```

It is specifically the `codeload.github.com` host that is blocked for repositories outside the session's configured set. A public control confirms codeload itself is reachable:

| request | result |
| --- | --- |
| `codeload.github.com/localthought/devonian/tar.gz/<sha>` | 403 |
| `codeload.github.com/ontola/atomic-server/tar.gz/refs/heads/develop` | 200 |

## Impact

The failure lands at the very end of a ~2 minute install, after all 1985 other packages have resolved and downloaded, and it is fatal: no `node_modules` for any of the 11 workspace projects. That means in these environments nobody can run `pnpm typecheck`, `pnpm lint`, the vitest suites or the Playwright e2e tests at all. Frontend changes get pushed with CI as the first real typecheck.

This has come up repeatedly in agent sessions on this repo, and the workarounds people reach for are bad ones: stripping the dependency out of `package.json` to install without it, which desyncs the lockfile, or skipping local verification entirely.

[...]

PR #1580, opened 2026-09-19
Format the tests that came with the stale-auth-proof fix

joepio:

Before: `cargo fmt --all --check` fails on develop's head. The pipeline runs that gate first, so it dies about twenty seconds in, before a single test starts. Develop produces no test information at all, and so does every branch that merges develop: run 4180 on a feature branch already failed exactly this way, with a 45-minute pipeline that reported nothing.

After: the gate passes, and develop's runs get far enough to tell anyone whether the code works.

The three offending hunks all came in with #1578, in tests: the `sign_message` call in `signed_at` and the `assert_eq!` on `error.error_type` in `lib/src/authentication.rs`, and the `get_client_agent` binding in `a_cookie_whose_proof_aged_out_is_the_public_agent` in `server/src/helpers.rs`. Nothing caught them because #1578's own develop run was concurrency-held behind the 19:17 README push and then cancelled before it ever ran. `main.yml` sets `cancel-in-progress: false` on develop, so develop runs strictly serialize and a stale run parks every later one.

## How

`cargo fmt --all` and nothing else. The diff is line wrapping plus one trailing comma after `&pair.private`, which is inert. No behaviour change, no test changed in what it asserts.

## Related Issues

None filed. Found while chasing why develop had no verdict.

## Checklist
- [ ] Update docs if needed

No changelog entry: this is formatting of unreleased test code, with nothing for a reader of the changelog to act on. No docs change for the same reason.

Verified with the repo's pinned 1.98.1 toolchain: `cargo fmt --all --check` fails on `71d36bf` in exactly these three places and passes on this branch. `cargo check -p atomic_lib --lib --tests` is clean. I could not run the full workspace check in this environment, because the server build script needs the `devonian` dependency that `pnpm` cannot fetch here (issue #1569), which is also why the pre-commit hook could not run.


joepio:
CI is red here on `ci --host-profile mancave --playwright-mode light`, and it is not this PR's.

The failing check is the light (`@smoke`) Playwright shard, in the chatroom spec. A message is typed and sent, and it never renders: the page still reads "No messages yet" and the send button is back to disabled, so `expect(page.locator('text=My test: 12:24:31 AM')).toBeVisible()` times out at 15s, on the original and both retries.

This PR changes two files, both Rust, both inside `#[cfg(test)] mod test` blocks, and the change is `cargo fmt --all` output: line wrapping plus one trailing comma. It compiles to the same program. Nothing in it can reach a browser test that sends a chat message.

What it does show is that the change works. The format gate is the first thing the pipeline runs, and on develop's head it fails in about twenty seconds, before any test starts. This run got through test, lint and build and spent twenty minutes in e2e, which is further than develop itself currently gets.

I have not ported a fix, because none exists yet and I do not yet know whether the chatroom failure is pre-existing or new. Develop has no concluded run on `71d36bf` to compare against: run 4178 is still pending behind the queue. The last develop run that concluded is 4160 on `59b3c00`, which predates both #1578 and #1579, and its failures were in the integration-visibility and plugin specs. So I am reproducing the chatroom spec locally against this base to find out whether develop regressed here, and will report separately. I have not spent the re-run on this yet.

---

joepio:
Reproduced, as promised above. The chatroom failure is develop's, and it is fully reproducible off CI.

I ran the spec on this branch (`2730434`, which is develop's head plus `cargo fmt` whitespace) on my own machine, against a wiped `.e2e-store`, one worker, no Dagger and no shared runner:

```
✘ [chromium] tests/e2e.spec.ts:291:7 › data-browser › chatroom @smoke (43.0s)
   at e2e.spec.ts:423 — page2.locator('text=My test: 12:32:05 AM') not visible
```

So it is not CI load, not store growth, and not this PR. That settles the question I left open: the failure is red on the base, and I have not needed the re-run.

**It is not #1578.** The stale-auth-proof fix was the leading suspect for develop's e2e redness. It is not what fails here. In the server log for this run, the caller on every refused read is a real authenticated agent, `did:ad:agent:rD9ZtSnZ…`, and `ForAgent::Public` appears zero times, so the `_or_public` fallback that PR introduced is never taken.

~~**What actually refuses the read** is `lib/src/hierarchy.rs:475` … the invite acceptance not granting the accepting agent read.~~

**[Edit: retracted. That lead does not hold, and I should not have stated it.]** I checked the log again against the test's timeline and actors:

- Only **one** agent DID appears in the whole run, `did:ad:agent:rD9ZtSnZ…`, first seen at 00:32:16. The message is sent at 00:32:05 and the test runs 43s, so 00:32:16 is the **owner's** phase, before `page2`'s context exists.
- **`page2`'s agent never appears in a refused rights walk at all.**
- The six `rights walk TERMINATED` lines span 00:32:16 to 00:32:28, straddling the owner phase — where the owner's own assertion at line ~328 **passed**. A denial that does not stop the owner seeing the message is not what stopped page2 seeing it.

So those warnings are most likely routine noise from a walk over a parentless subject, and the invite-rights theory should not be the starting point for anyone picking this up.

What stands: the spec fails locally on develop's head with a wiped store and no CI, so it is develop's own and not this PR's; and #1578's public-agent fallback is genuinely not exercised. What is unknown: why page2 never sees the owner's message.

Separately, on this PR's own validation: I have now run the pre-commit hook's exact command in a checkout where `browser/data-browser/dist` exists, rather than relying on the hook, which cannot run in my environment:

```
cargo clippy --workspace --exclude atomic-server-tauri --no-deps --all-targets \
  --no-default-features --features light,wasm-plugins -- -D warnings
```

None of the chatroom investigation is this PR's to fix, and this PR does not make it worse. The format gate is still what blocks every develop run from reporting at all, which is what this one commit of whitespace is for.

PR #1589, opened 2026-09-20
Guard the slash menu against positioning a destroyed renderer

joepio:

Before: typing `/` in the rich text editor opens the command list, and pressing Escape within the first animation frame throws `Cannot read properties of undefined (reading 'element')`. The menu closes either way, so what the throw costs is the rest of that frame's work rather than the dismissal itself.

After: Escape closes the menu with nothing thrown, whenever it is pressed.

`onStart` positions the list and then schedules a second positioning pass on the next animation frame. Escape in that window runs `destroyComponent`, which clears the renderer reference that the scheduled pass goes on to read.

## How

The check moves into `updatePosition`, where the reference is actually read, instead of being repeated at each call site. `onUpdate` and `onKeyDown` already bail on a missing renderer; the deferred call from `onStart` was the one path that did not.

Worth knowing for reviewers: this was also the only `tsc --noEmit` error in `data-browser` on develop (`CommandsExtension.ts(72,23): error TS18048`). It went unnoticed because `.dagger/src/index.ts` has no typecheck step, so CI never runs `pnpm typecheck` at all. The four sibling errors in `browser/e2e` were fixed separately in `0bce071`, so once this merges the workspace is clean and the step can be added.

## Verification

CI cannot validate this at the moment: `Main (Mancave) / CI` dies at `Log in to Docker Hub` on the runner host, as the comment below sets out. Run locally instead, against the vite dev server so the change is actually the code under test:

- `tsc --noEmit -p tsconfig.json` in `browser/data-browser`: clean.
- `website-inline-rte.spec.ts`, which types `/quo`, asserts the command list, presses Escape, asserts it hidden, then reopens and accepts an entry: 1 passed in 26.9s.

## Related Issues

None.

## Checklist
- [ ] Update docs if needed


joepio:
`Main (Mancave) / CI` is red here, and it is not this branch.

The job died 15 seconds in at step 4, `Log in to Docker Hub`, and steps 5 through 10 including the Dagger step are all `skipped`, so nothing in this diff was ever built or run. `Downstream (atomic-saas builds against this commit)` passed on the same runner in the same run, taking 7 and a half minutes.

The cause is on the runner host, not in any commit. A plain bash step in atomic-saas CI that calls docker without Node anywhere in the path gets:

```
/usr/bin/docker: Input/output error
```

with exit code 126, which is "found but not executable". The `docker/login-action` failure here is the same fault surfacing through Node as `Error: spawn EIO`, and the post-job `docker logout` fails identically. `/usr/bin/git version` runs fine in the same job seconds later, so it is that one binary rather than the machine.

Seven `Main (Mancave) / CI` jobs have died this way since 14:27 today, including develop's own runs 4210 and 4219 on commits this branch does not contain. There is no fix to port into this PR, because the breakage is host state rather than a change.

I am not spending a re-run on it. The failure is deterministic and has been reproduced seven times in three hours plus once at the shell, so a re-run adds nothing and costs the queue another slot on a runner that is already the bottleneck. This PR stays watched and I will say so here when a pipeline gets past that step.

---

joepio:
Superseded and closing. The same guard landed on develop directly as `195737d0`, with an equivalent changelog entry, so there is nothing left in this branch to merge. The code is character for character the same check in `updatePosition`; only the comment above it differs.

For the record, since the reasoning is worth keeping with the fix: `onStart` positions the command list and schedules a second pass on the next animation frame, and Escape in that window runs `destroyComponent`, so the scheduled pass read `element` off a cleared reference. It was also the last `tsc --noEmit` error in `data-browser`, which no CI job runs today because `.dagger/src/index.ts` has no typecheck step. That gap is still open and is being handled separately.

---

PR #1595, opened 2026-09-21
Let the integration proxy live on a .localhost name

joepio:

Before: opening Settings in CI showed "Error loading resource" with `Proxy must be an HTTPS origin or localhost`, and the page never rendered. Every spec that starts by searching Settings then timed out looking for a search box that was not there. In develop's run 4247 that accounted for at least a third of the failures, `settings.spec.ts` among them, which has nothing to do with integrations.

After: Settings renders, and a proxy on a `.localhost` name is accepted the way a browser already treats one.

CI builds the e2e bundle with `VITE_INTEGRATION_PROXY_URL=http://atomic.localhost:19090`, and it has to: the browser runs in the Playwright container, where `127.0.0.1` is that container rather than the machine hosting the mock proxy. `proxyOrigin` allowed plain http only for the bare names `localhost` and `127.0.0.1`, so it threw on exactly the value CI bakes in. Running the suite locally never saw it, because `e2e-server.sh` points the bundle at `http://127.0.0.1:<port>` instead.

## How

`proxyOrigin` now treats any `.localhost` name as loopback, `[::1]` with it. RFC 6761 gives the whole `.localhost` TLD to loopback and browsers grant those names secure-origin status for that reason, so this is what the check meant to say.

The throw mattered as much as the value did. The proxy is read during render in eight places across seven components, either through `useIntegrationProxy()` (`IntegrationSettings.tsx:20`, `localThoughtCallback.tsx:176`, `IntegrationDiscovery.tsx:139`, `LocalThoughtCatalog.tsx:20`, `ConnectNotion.tsx:39`) or as a bare `getIntegrationProxy()` in a default argument or lazy initializer (`ConnectLocalThought.tsx:24`, `localThought.ts:34` and `:90`, `DevonianDemoRoute.tsx:29`). Every one of those reaches `proxyOrigin` through `configuredProxy`, with nothing to catch it, so any route mounting any of them died whole rather than degrading to one broken integration. A captured page snapshot puts the throw at `DevonianDemoRoute.tsx:29`'s `useState(getIntegrationProxy)`, through `renderWithHooks`.

`configuredProxy` also falls back to the baked-in value whenever `localStorage` has no `integration-proxy-url` key, which is every fresh e2e context, so the bad value was reached on a first visit rather than only after someone configured something.

## Related Issues

None; found while reading develop's run 4247.

## Checklist
- [ ] Update docs if needed

Unit tests sit beside the existing `proxyOrigin` cases in `integrations/localthought/browser.test.ts` and cover the CI value, the bare names, and two rejections that must stay rejections: `http://notlocalhost` and `http://localhost.evil.example`. The 62 tests of that suite pass locally. No docs mention the restriction.


joepio:
Merged with CI red, on purpose, so here is the measurement behind that.

A full e2e run on this branch ([4255](https://github.com/ontola/atomic-server/actions/runs/35575570247)) against develop's own full run on `20fe0d7` ([4247 attempt 2](https://github.com/ontola/atomic-server/actions/runs/35537851934)), same host profile, same four shards:

| | 4247 (develop) | 4255 (this branch) |
|---|---|---|
| failed | 34 | 14 |
| flaky | 2 | 4 |
| passed | 244 | 262 |
| skipped | 11 | 11 |

Twenty-two of develop's 34 cleared, and `Proxy must be an HTTPS origin or localhost` appears zero times in 4255's log against 126 occurrences of the downstream `Search settings...` timeout in 4247's. Cleared: `app-setup` :10 :41 :79, `devonian-issue-sync` :9 :252, `integration-store-install`:14, `integration-visibility`:6, `mt940`:16, `ollama-feedback`:24, `plugins` :30 :109 :163 :394 :523 :889 :935 :1623 :1676 :1740 :1774, `settings`:7, and `table-refresh`:26 down to flaky.

The 14 that remain are develop's, not this change's. Nothing here touches them: the diff relaxes one validator and adds tests. Two are already fixed on a branch waiting to land (`apps.spec.ts`:22, and `drive-template-onboarding`:228 with `new-resource-catalog` :73 :267, which #1593 broke by renaming the AI setup panel title). Four are the known contention rotation on this host, reproduced off the box the same day: the same files under load fail a different subset each run. `plugin.spec.ts`:26 is still unexplained and now has CI data for the first time.

Two failures in 4255 were not in 4247 (`timer.spec.ts`:156, and `installation-recovery`:96 moving from flaky to failed) and one flaky is new (`template.spec.ts`:293). None involves integrations, and `template.spec.ts`:293 is the documented two-workers case. That is the rotation, not this diff.

---

PR #1596, opened 2026-09-21
ci: temporarily publish Docker images from feat/plugin-debug

michielbdejong:
## Related Issues

None — requested directly by @michielbdejong to speed up `localthought/atomic-plugins`' CI.

## Summary

`localthought/atomic-plugins` (a temporary repo iterating on `integrations/` before it lands back here) pins its own CI to an exact commit on this branch via `.atomic-server-ref`, and builds `atomic-server` from source on every run to get a server to test against — there's currently no published image to pull, since this workflow only publishes from `develop`. That's especially costly in cold/ephemeral containers (e.g. Claude Code cloud sessions) where nothing is cached between runs.

This adds:
- A full-commit-SHA tag (`type=sha,format=long`) alongside the existing default tag preset in the `docker/metadata-action` step, so `atomic-plugins` can pull `ghcr.io/ontola/atomic-server:<the exact SHA in .atomic-server-ref>` without needing to compute/query a shorter tag.
- A `|| github.ref == 'refs/heads/feat/plugin-debug'` clause on the existing "Dagger docker images (build images & publish)" step, alongside the existing `develop` publish.

**Temporary, by design** — the PR comment in the diff says so, and it should be reverted once the plugin interface stabilizes and `atomic-plugins` starts integrating against `develop` instead of this branch.

## Checklist
- [x] No changelog entry — CI-only, no API/product change
- [ ] No tests added — nothing here is testable beyond the workflow itself running (verifiable once merged: the next `feat/plugin-debug` push should publish `ghcr.io/ontola/atomic-server:<sha>`)
- [x] This PR body documents the change and its intended removal

https://claude.ai/code/session_015EAuJBtDDn9Wfy3TnZ2vbY

---
_Generated by [Claude Code](https://claude.ai/code/session_015EAuJBtDDn9Wfy3TnZ2vbY)_


michielbdejong:
The `ci --netlify-auth-token env://NETLIFY_TOKEN --host-profile mancave --playwright-mode light` check is red on this PR, but it's not this PR's failure: `858103e` — this PR's base commit and the current HEAD of `feat/plugin-debug` — already failed with the exact same step (`Dagger CI (test, lint, build, e2e light)`) in its own push-triggered run before this PR existed ([run 35372103892](https://github.com/ontola/atomic-server/actions/runs/35372103892/job/105744789389)). This PR's diff only touches the metadata-extraction and docker-publish steps that run *after* `dagger call ci`, so it can't be the cause.

I don't have a fix for the underlying `feat/plugin-debug` failure — it's outside the scope of this change (a Docker-publish CI gate), and I haven't root-caused the broader Dagger test/lint/build/e2e failure. Not spending the one re-run here since the base branch's own independent run already reproduces it identically, which is confirmation enough. Leaving this PR watched.

---

michielbdejong:
Keeping this one up my sleeve in case I need the speedup

michielbdejong:
The `Downstream (atomic-saas builds against this commit)` check also failed on this run, and unlike the Dagger CI check, it *passed* on the base commit's own prior run — so I can't wave it off the same way. Looked at the log: it's `atomic-saas`'s current `main` (rebuilt fresh each run against whichever `atomic-server` commit triggers it) failing with `E0599: no method named managed_alias_hosts found for &atomic_lib::Db` and the same for `sync_drive_mappings`. Nothing in this PR touches `lib/` or any Rust source — the diff is one CI workflow YAML file — so this reads as `atomic-saas@main` having drifted ahead of `atomic_lib`'s current API surface in the few days since the base commit's last run, not something this PR introduced or can fix from here (I don't have `atomic-saas` in scope, and it isn't this branch's job to carry that API forward).

Between this and the pre-existing Dagger CI failure noted above, and the PR now being a draft, I'm leaving both as known-red rather than chasing further — flag if you want me to dig into either.

---

michielbdejong:
let's not do this, we can get the speedup when things land in develop, and until then it will have to rely on build cache

